PR #6, opened 2022-02-02
'Solution'

serhii56465:


denyslipin (approved):

app/main.py
```diff
@@ -1 +1,32 @@
...
+            adpath += i
+            os.mkdir(adpath)
+            adpath += "/"
+        with open(sourcefile) as f:
+            lines = f.readlines()
+        if len(destination[-1]) != 0:
+            with open(f"{'/'.join(directory)}/{copyfile}", "a") as f1:
+                f1.writelines(lines)
+        else:
+            with open(f"{'/'.join(directory)}/{sourcefile}", "a") as f1:
```
MasterpieceElbow, line 29: istead of `'/'.join` use `os.path.join` because on different operating systems can be different connectors.

MasterpieceElbow (changes requested):

IvanRamyk (approved):

PR #11, opened 2022-05-28
Solution

rotsen18:


app/main.py
```diff
@@ -2,7 +2,7 @@
 
 
 def move_file(command: str):
-    operator, current_file_name, destination = command.split()
+    operator, old_file_name, destination = command.split()
     *folders, new_file_name = destination.split("/")
 
     if operator != "mv":
```
Serhii-Leonenko, line 8: In my opinion, it's better to raise an error here.

Everything else is good !

Serhii-Leonenko (approved):

MasterpieceElbow (approved):

app/main.py
MasterpieceElbow: Yes, raising an error is better here but it is unnecessary in this task.

PR #12, opened 2022-05-29
'Solution'

Korenely:


DmytroSvirsa (changes requested):

app/main.py
```diff
@@ -1 +1,19 @@
...
+import os
+
+
+def move_file(command):
+    sourse = command.split()[1]
+    target = command.split()[2]
+
+    if "/" in target:
+        target_split = target.split("/")
+        target_path = os.path.join(*target_split)
```
TPolina, line 10: you split the `target` and then join it, so you get exactly what you had without any difference

app/main.py
```diff
@@ -1 +1,19 @@
...
+
+        os.makedirs("/".join(target.split("/")[:-1]))
+
+        with open(target_path, "w") as f:
+            f.write(open(sourse, "r").read())
```
TPolina, line 15: not a good way to open file, better use the second context manager

TPolina (changes requested):

TPolina (approved):

PR #14, opened 2022-05-29
All done.

Flashmobber:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+import os
+
+
+def move_file(command: str):
+    parsed = command.split()
+
+    if parsed[0] == "mv" and parsed[1] != parsed[2]:
```
rotsen18, line 7: I would use a few variables instead "parsed[0]" and other in all code.

rotsen18, line 7: I would use a few variables instead "parsed[0]" and other in all code.

app/main.py
```diff
@@ -1 +1,20 @@
...
+        if os.path.isfile(parsed[2]):
+            return
+
+        if parsed[2][-1] == os.path.sep:
+            parsed[2] += parsed[1]
+
+        path = ""
+        for d in parsed[2].split(os.path.sep)[:-1]:
```
rotsen18, line 15: Try to use more descriptive names

rotsen18, line 15: Try to use more descriptive names

rotsen18 (approved):

app/main.py
MasterpieceElbow, line 7: If you use `parsed[0]` / `parsed[1]` /  `parsed[2]` several times - save it to the variables.

app/main.py
MasterpieceElbow, line 15: `d` is not a good variable name.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #16, opened 2022-05-30
Solution

YuriiKindrat:


rotsen18 (approved):

TPolina (approved):
approve, but better don't write a lot of code inside `with` block.

PR #18, opened 2022-05-31
solution

BohdanKryven:


rotsen18 (approved):

app/main.py
```diff
@@ -1 +1,29 @@
...
+import os
+
+
+def move_file(command: str):
+    command = command.split()
+    command_values = {
+        "mode": command[0],
+        "file": command[1],
+        "file_move": command[2]
+    }
```
TPolina, line 10: what for you use dictionary here? You can just create 3 variables and use them without complication

TPolina (changes requested):

PR #28, opened 2022-06-10
Solution

pavlejviki:


Abnormaltype (changes requested):
Good job!

But this command is not working `move_file("mv file.txt new_file.txt")`

PR #29, opened 2022-06-19
Solution

Splend1ed:


app/main.py
```diff
@@ -1 +1,22 @@
...
+    path_to_move = command.split()[2]
+
+    old_file = open(old_file_name, "r")
+    dirs = path_to_move.split("/")[:-1]
+    path_without_file = "/".join(dirs)
+
+    if len(path_to_move.split("/")) > 1:
+        makedirs(path_without_file)
+
+    new_file = open(path_to_move, "w")
```
IvanRamyk, line 15: Let's use context managers
It's a more common way to work with files

IvanRamyk (changes requested):

PR #30, opened 2022-06-22
Solution

artemkazakov947:
Sorry for wierd code.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -2,14 +2,14 @@ import os
 
 
 def move_file(command):
-    origin = command.split(" ")[1]  # source file_name
-    if "/" not in command:  # check if we need just rename file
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,24 @@
-# write your code here
+import os
+
+
+def move_file(command):
+    origin = command.split(" ")[1]  # source file_name
```
IvanRamyk, line 5:
```suggestion
    origin = command.split(" ")[1]
```
We believe that code should be self-descriptive
So you don't need to write comments

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #51, opened 2022-08-18
'Solution'

Viktor-Beniukh:


DmytroSvirsa (changes requested):
also, you should check the input command

app/main.py
```diff
@@ -3,9 +3,13 @@
 
 def move_file(command):
 
+    command = command.split()
     file_curr = command.split()[1]
     file_curr_move = command.split("/")[-1]
 
+    if command[0] != "mv":
+        print("Command entered incorrectly!")
+
```
DmytroSvirsa, line 12: ok, you find an error, print the message but don't interrupt and allow to do all following steps

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #54, opened 2022-08-21
Solution

volodymyr-vereshchak:


Abnormaltype (changes requested):
use `""` instead of `''`

Abnormaltype (approved):
Well done!

PR #67, opened 2022-09-19
'Solution'

davidkrivko:


app/main.py
```diff
@@ -1 +1,20 @@
...
+    values = values.split()
+    path = values[2].split("/")
+    with open(values[1], "r") as file_in:
+        content = file_in.read()
+    os.remove(values[1])
+    if len(path) == 2:
+        os.mkdir(path[0])
+    if len(path) == 3:
+        os.mkdir(path[0])
+        os.mkdir(path[0] + "/" + path[1])
```
IvanRamyk, line 14: Use `os.path.join` to concatenate parts of the path

app/main.py
```diff
@@ -1 +1,20 @@
...
+    if len(path) == 4:
```
IvanRamyk, line 15: What if 5 or more?

app/main.py
```diff
@@ -1 +1,20 @@
...
+        os.mkdir(path[0])
```
IvanRamyk, line 16: if the directory already exists then it will lead to an error

IvanRamyk (changes requested):

app/main.py
davidkrivko: thank you

app/main.py
```diff
@@ -1 +1,22 @@
...
+    with open(values[1], "r") as file_in:
+        content = file_in.read()
+    os.remove(values[1])
+
+    path_1 = ""
+    for i in path_:
+        if i == path_[-1]:
+            path_1 += os.path.join(i)
+        path_1 += os.path.join(i + "/")
+        os.mkdir(path_1)
```
YuriiKindrat, line 16: what is path and i variables mean ? make a varables name more informative

YuriiKindrat (changes requested):

PR #68, opened 2022-09-19
Solution

yevhenii-nevmyvako:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+import os
+
+
+def move_file(command):
+    split_file = command.split(" ")
+    if "/" in split_file[-1]:
+        path = split_file[-1].split("/")
+        for i in range(len(path) - 1):
+            os.mkdir("/".join(path[:i]))
```
IvanRamyk, line 9: Code fails in this line

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+import os
+
+
+def move_file(command):
+    split_file = command.split(" ")
+    if "/" in split_file[-1]:
+        path = split_file[-1].split("/")
+        os.makedirs("/".join(path[:-1]))
```
DmytroSvirsa, line 8: Use `os.path.join` to concatenate parts of the path. 
In that case, your app will be cross-platform and will 
use either `/` or `\` depending on OS

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+import os
+
+
+def move_file(command):
+    split_file = command.split(" ")
```
DmytroSvirsa, line 5: Make sure you check that the `mv` input command is correct.

app/main.py
DmytroSvirsa, line 8: Notice that if you will try to create the directory already exists then it 
will lead to an error.

DmytroSvirsa (changes requested):

app/main.py
DmytroSvirsa: Make sure you check that the `mv` input command is correct.

DmytroSvirsa (changes requested):

DmytroSvirsa (changes requested):
please read the comments above (if something is unclear, feel free to ask the questions)

DmytroSvirsa (approved):

PR #72, opened 2022-09-21
Solution

AriohBelskij:


wQuelS (approved):
Looks good, but idk about "/".join()
I'd better used os.path.join.

Nattalli (approved):
Good job! 😸

PR #77, opened 2022-09-25
'Solution'

AVyha:


mltdll (changes requested):
Please provide names for your command arguments, don't just use `command` as an array. It makes your code hard to read and understand.

mltdll (approved):

Nattalli (approved):
Good job!

PR #79, opened 2022-09-26
Solution

StasOkhrym:


mltdll (approved):

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from os import remove, rename, makedirs, path
+
+
+def move_file(command: str) -> None:
+    command = command.split()[1::]
+    if "/" in command:
+        file_path, file_name = command.rsplit("/", 1)
```
Nattalli, line 7: does it work fine for you? Because I have this problem:
`
AttributeError: 'list' object has no attribute 'split'
`

Nattalli (changes requested):

app/main.py
StasOkhrym: sorry, kind of typo in `command.split()[1::]`

app/main.py
```diff
@@ -1 +1,17 @@
...
+        makedirs(file_path)
+    else:
+        new_file_name = command_parts[-1]
+        rename(file_name, new_file_name)
+        return
+    with open(file_name, "r") as source, \
+            open(path.join(file_path, file_name), "w") as destination:
+        destination.write(source.read())
+        remove(file_name)
```
Nattalli, line 17: but it doesn't remove the file

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from os import remove, rename, makedirs, path
+
+
+def move_file(command: str) -> None:
+    command_parts = command.split()
+    file_name = command_parts[1]
+    if "/" in command:
+        file_path, file_name = command.rsplit("/", 1)
+        makedirs(file_path)
```
Nattalli, line 9: also, it makes not correct path:
1) `mv` at the created directory name
2) new file doesn't create

Nattalli (changes requested):

Nattalli (approved):
well done!

PR #81, opened 2022-09-28
Solution

taras-andruschenko:


app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command = command.split()[1:]
```
mltdll, line 5: I'd think you should introduce variables for each argument so that code is easier to read.

app/main.py
```diff
@@ -1 +1,19 @@
...
+    if "/" not in command[1]:
+        os.rename(command[0], command[1])
+    else:
+        path = command[1].split('/')
+        true_path = ""
+        for i in range(len(path) - 1):
+            true_path += path[i]
+            os.mkdir(true_path)
```
mltdll, line 13: You can also use `os.mkdirs()` function instead of writing loop yourself. Though README does actually suggest to use `mkdir`. :)

mltdll (changes requested):

MasterpieceElbow (approved):
GJ!

PR #87, opened 2022-09-29
Solution)

Gr1zzZzzly:


app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command = command.split()
```
mltdll, line 5: I think you should introduce separate variables for both arguments. It's hard to follow your code's logic when they are just `command[1]` and `command[2]`.

mltdll (changes requested):

MasterpieceElbow (approved):
GJ!

PR #88, opened 2022-09-30
Solution

pazyn404:


mltdll (approved):
Judging by the amount of stuff you imported from `os`, it seems to me that you've deliberately done by hand what existing solutions such as `os.path.exists()` and `os.makedirs()` do.

Nattalli (approved):
Looks good 😄

PR #89, opened 2022-10-01
Solution

rostyak:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+import os
+import shutil
+
+
+def move_file(command: str) -> None:
+    if command.split()[0] == "mv" and not command.endswith("/"):
```
mltdll, line 6: It is usually considered best practice not to put function's main logic into an `if` condition. I'd suggest to rewrite it like this:
```python
if ...:
    return

...

mltdll (changes requested):

app/main.py
rostyak: I wanted to do this, but I thought it would be wrong

mltdll (approved):

app/main.py
mltdll: Look at your code now: it's immediately clear that the function terminates when the `if` block executes, so the reader doesn't have to look any further to double check.

Aside from that, there was an extra layer of indentation that is now gone. This is also pretty good. :)

Y-Havryliv (approved):
GJ!

PR #91, opened 2022-10-01
Solution

Andriy-Makarenko:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from os import makedirs, remove
+
+
+def move_file(command: str):
+    command_parts = command.split()
+    source_file = command_parts[1]
+    second_file_path = command_parts[2]
+    destination_folder = second_file_path.rsplit("/", 1)[0]
+    makedirs(destination_folder)
```
mltdll, line 9: This function will produce an error if destination directory already exists. It can be changed with `exists_ok` parameter.

app/main.py
```diff
@@ -1 +1,17 @@
...
+
+    with open(source_file, "r") as file:
```
mltdll, line 11: You can also open both files at once, but I don't think it's necessary.

mltdll (changes requested):

mltdll (approved):

TPolina (approved):
Well done!

PR #92, opened 2022-10-01
Complete!

ketstap162:


app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command = command.split()
+    if command[0] != "mv":
+        return None
```
mltdll, line 7: When return is used just to terminate function, it is usually written as plain `return`, not `return None`. Not sure if that's an issue, though.

mltdll (approved):

Y-Havryliv (approved):
GJ!

PR #93, opened 2022-10-02
Solution

luxuriant777:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+
+
+def move_file(command):
+    if command.split()[0] == "mv":
+        file = command.split()[1]
+        path = command.split()[2]
```
TPolina, line 7: you use `split` 3 times while you can use it once

app/main.py
```diff
@@ -1 +1,17 @@
...
+        new_file = path.split("/")[-1]
+        path = path.replace(new_file, "")
+        new_path = os.path.join(path)
+        os.makedirs(new_path)
+        with open(file, "r") as source_file:
+            file_copy = source_file.read()
+        os.remove(file)
+        os.chdir(new_path)
+        with open(new_file, "w") as new_file:
+            new_file.write(file_copy)
```
TPolina, line 17: too much code in if block but it's not necessary here. Can you refactor it?

TPolina (changes requested):

app/main.py
luxuriant777: changed

TPolina (approved):
Good job!

PR #95, opened 2022-10-03
'Solution'

hasslieb3:


app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+from os import mkdir, remove
+
+
+def move_file(command: str) -> None:
+    command = command.split()
+    original_file_name = command[1]
+
+    with open(original_file_name, "r") as file:
```
HStakhiv, line 8: You can open both files at the same time

app/main.py
```diff
@@ -1 +1,19 @@
...
+        data = file.read()
+
+    path = ''
+    for folder in command[2].split('/')[:-1]:
+        path += f"{folder}/"
+        mkdir(path)
+
+    with open(f"{command[-1]}", "w") as copied_file:
```
HStakhiv, line 16: You can open both files at the same time. Read and write information simultaneously

HStakhiv:
```python
with open(..., "r") as file_in, open(..., "w") as file_out:
    ...
```

app/main.py
```diff
@@ -1 +1,19 @@
...
+
+def move_file(command: str) -> None:
+    command = command.split()
+    original_file_name = command[1]
+
+    with open(original_file_name, "r") as file:
+        data = file.read()
+
+    path = ''
+    for folder in command[2].split('/')[:-1]:
```
HStakhiv, line 12: What if we don't have such a symbol (`/`)? Probably better to check for a `/` symbol in the second name

HStakhiv (changes requested):
The function should move the file only if the command is `mv`.

HStakhiv (approved):

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command = command.split()
+    first_file = command[1]
+    second_file = command[2]
+    if len(os.path.dirname(second_file)) > 0:
```
TPolina, line 8: strange check. Why did you decide to use `len` here? It's not obvious for me

app/main.py
```diff
@@ -1 +1,14 @@
...
+        os.makedirs(os.path.dirname(command[2]), exist_ok=True)
```
TPolina, line 9: Why do you use command[2] when you have second_file?

TPolina: too many `dirname` methods in your code

app/main.py
```diff
@@ -1 +1,14 @@
...
+        with open(second_file, "w") as f:
+            f.write("")
```
TPolina, line 11: Where do you write content of origin file to source file?

app/main.py
```diff
@@ -1 +1,14 @@
...
+        os.remove(os.path.dirname(second_file) + "\\" + first_file)
```
TPolina, line 12: firstly, created path will not work for some OS
secondly, what do you do here at all?

TPolina (changes requested):
Did you test your task? I see a mess here

app/main.py
hasslieb3: rewritten again

TPolina (approved):
Well done!

PR #99, opened 2022-10-03
Solution

VadymShkarbul:


app/main.py
```diff
@@ -1 +1,22 @@
...
+        return f"{cmd[0]} not a move command"
+
+    path = cmd[2].split("/")
+    new_file = path.pop()
+    path = "/".join(path)
+
+    if not os.path.isdir(path):
+        os.makedirs(path)
+    try:
+        with open(cmd[1], "r") as source, \
```
mltdll, line 16: I think it's better to use parentheses here.
```python
with (open...,
      open...):

**Review by mltdll — CHANGES_REQUESTED** [review]

File: app/main.py [hunk 2]
```diff
@@ -1 +1,22 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> str | None:
+    cmd = command.split()
+    if cmd[0] != "mv":
+        return f"{cmd[0]} not a move command"
```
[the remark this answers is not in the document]
VadymShkarbul: Thank you for the review! As for me, this function looks like a "under the hood function", but which works with the user's request, and it would be nice not to just not fulfill the request, but to explain why. Agree that these solutions do not cover all possible input errors. And on 7th line, I did not want to make this decision more complicated and write an additional custom exception. "Simple is better than complex", you know ;)

app/main.py
VadymShkarbul: I didn't understand what you mean here. I use "\" just for splitting

app/main.py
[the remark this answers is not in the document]
mltdll: Okay, seems like a valid reasoning. What about my other comment? Do you want to change anything or do I just approve?

app/main.py
VadymShkarbul: Aaa understand! Thank you! Give me 5 minutes to change. It's cool!

mltdll: Oops, for some reason your reply didn't load when I opened the page last time. Good thing you understood yourself! :)

VadymShkarbul: It seemed that you were proposing to shorten the code due to fewer brackets, but I tried it and realized that it's not - please approve ;) In general, I probably need to be more sure in myself. Thank you!

mltdll: Yes, this would not shorten the code. It would add one more symbol to it.
However, the mentors at Mate seem to be much more fond of using parentheses than backslashes for line breaks. It also infected me.

mltdll (approved):

Y-Havryliv (approved):
GJ!

PR #100, opened 2022-10-03
'Solution'

serviza:


alexzubkoff (commented):
Looks good! Nothing to add.

app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+import os
+
+
+def movie_file(command: str):
+    command_list = command.split()
+    if command_list[0] != 'mv':
```
Y-Havryliv, line 6: use double quotes everywhere

Y-Havryliv (changes requested):

Y-Havryliv (approved):
GJ!

app/main.py
serviza: To be the same everywhere? Or is there an advantage over single quotes?

PR #101, opened 2022-10-03
Solution

FranchukS:


app/main.py
```diff
@@ -1 +1,24 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command = command.split()
```
mltdll, line 5: I think it would be much better to create variables for both command arguments, since it would be easier to follow your code logic with named variables than with `command[0/1/2]`.

app/main.py
```diff
@@ -1 +1,24 @@
...
+    if command[0] == "mv" and command[2][-1] != "/":
```
mltdll, line 6: Also, I think it's not a great idea to put your whole function into an `if` block. I would suggest to reformat it like this:
```python
if ...:
    return
...

mltdll (approved):

app/main.py
mltdll: And here's graphical representation of the problems this style presents, in my opinion. :)
image

FranchukS: Good point

app/main.py
FranchukS: make sense

mltdll (approved):
Oh, did accidentally approve this PR right away? Either way, GJ.

TPolina (approved):
Good job!

PR #102, opened 2022-10-04
Solution

misha-doroshenko:


app/main.py
```diff
@@ -1 +1,39 @@
...
+    for i in range(len(directories) - 1):
+        full_dir += directories[i] + "/"
+        if not os.path.exists(full_dir):
+            os.mkdir(full_dir)
+    full_dir += directories[len(directories) - 1]
+    return full_dir
+
+
+def move_file(command: str) -> None:
+    file_to_move = command.split(" ")[1]
```
FranchukS, line 29: The split method work with space (" ") by default setting, so you don`t have to point it

app/main.py
```diff
@@ -1 +1,39 @@
...
+    if not os.path.exists(file_to_move):
+        print(f"File '{file_to_move}' does not exist")
+        return
+
+    full_path = create_directories(command.split(" ")[2])
```
FranchukS, line 34: Same about the split method

app/main.py
```diff
@@ -1 +1,39 @@
...
+
+
+def create_directories(path: str) -> str:
+    directories = path.split("/")
+    full_dir = ""
+    for i in range(len(directories) - 1):
+        full_dir += directories[i] + "/"
+        if not os.path.exists(full_dir):
+            os.mkdir(full_dir)
+    full_dir += directories[len(directories) - 1]
```
FranchukS, line 24: You can use simpler index - to pick the last value of the array you can try [-1]

app/main.py
```diff
@@ -1 +1,39 @@
...
+    def __init__(self, file_name: str) -> None:
+        self.filename = file_name
+
+    def __enter__(self):
+        self.file = open(self.filename, "r")
+        return self.file
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if os.path.exists(self.filename):
+            os.remove(self.filename)
```
FranchukS, line 14: Test raise an Error in this place, it says:
with (ReadAndDelete(file_to_move) as f_read,
    os.remove(self.filename)
PermissionError: [WinError 32] Процесс не может получить доступ к файлу, так как этот файл занят другим процессом: 'Users.txt'

app/main.py
```diff
@@ -1 +1,39 @@
...
+class ReadAndDelete:
+    def __init__(self, file_name: str) -> None:
+        self.filename = file_name
+
+    def __enter__(self):
+        self.file = open(self.filename, "r")
+        return self.file
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if os.path.exists(self.filename):
```
FranchukS, line 13: I think you don`t need this "if " statement, because if this method runs - the file already exists because it`s a method of the file

FranchukS (changes requested):

app/main.py
misha-doroshenko: I didn't understand where the error should have occurred. Can you clarify this for me?

FranchukS (approved):
Well done!

TPolina (approved):
Unexpected and a bit weird solution through the context manager but ok)

PR #108, opened 2022-10-12
Solution of py-move-file

YLobunets:


app/main.py
```diff
@@ -1 +1,79 @@
...
+                return character_separator.join((self.full_path, filename))
+            return self.full_filename
+
+    command = user_command.split()
+    if len(command) != 3 or \
+            command[0].lower() != "mv" or \
+            command[1] == command[2]:
+        return
+
+    fileinfo1 = FileInfo(command[1])
```
IvanRamyk, line 40:
```suggestion
    source = FileInfo(command[1])
```
Not a really good name

app/main.py
```diff
@@ -1 +1,79 @@
...
+    if not os.path.isfile(fileinfo1.full_filename) \
+            or fileinfo1.path_only is True:
+        return
+
+    fileinfo2 = FileInfo(command[2])
```
IvanRamyk, line 45: same

IvanRamyk (changes requested):

app/main.py
YLobunets: Corrected

app/main.py
YLobunets: Corrected

app/main.py
```diff
@@ -1 +1,79 @@
...
+import os
+
+
+BUFFER_SIZE = 2048
+
+
+def move_file(user_command: str) -> None:
+    character_separator = "/"
+
+    class FileInfo:
```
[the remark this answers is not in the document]
YLobunets: The class belongs exclusively to the function, so it was done this way (inner class).
I had an idea - if in the future someone needs this function, then it is easier to copy it into other code.

Class declared at the top of the module.

IvanRamyk (approved):
Well done!

PR #110, opened 2022-10-17
'Solution'

Yurchikkk:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command = command.split(" ")
+
+    source_file = command[1]
+    path = command[2].split("/")
+    path_final = "/".join(path[:-1])
```
DmytroSvirsa, line 9: Use `os.path.join` to concatenate parts of the path. In that case, your app will be cross-platform and will use either `/` or `\` depending on OS

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command = command.split(" ")
+
+    source_file = command[1]
```
DmytroSvirsa, line 7: Make sure you check that the `mv` input command is correct.

DmytroSvirsa (approved):

DmytroSvirsa (changes requested):
read comments

PR #111, opened 2022-10-19
Solution

yessloss:


app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    linux_mv = command.split()[0]
+    first_file_name = command.split()[1]
+    directory_file = command.split()[2]
```
DmytroSvirsa, line 7: use unpacking to make this in 1 line

app/main.py
```diff
@@ -1 +1,13 @@
...
+    if linux_mv == "mv":
+        if "/" in directory_file:
+            os.makedirs(directory_file)
```
DmytroSvirsa, line 10: what about file?

DmytroSvirsa (changes requested):

app/main.py
yessloss: what with file?

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    file_name = command.split()
+    if file_name[0] == "mv":
+        if "/" in file_name[2]:
+            os.makedirs(file_name[2])
+        else:
```
DmytroSvirsa, line 9: you need to move file with all fufilness now you create file you need but original file is still exists and created file is empty, or I missing something?

DmytroSvirsa (changes requested):

Abnormaltype (approved):
Well done!

PR #116, opened 2022-10-30
Develop

Terrrya:


app/main.py
```diff
@@ -1 +1,20 @@
...
+        print("Please enter the right command")
+        return
+    if "/" in command_list[2]:
+        path_list = command_list[2].split("/")
+    if "\\" in command_list[2]:
+        path_list = command_list[2].split("\\")
+    path = ""
+    for i in range(len(path_list) - 1):
+        path = os.path.join(path, path_list[i])
+        os.mkdir(path)
```
IvanRamyk, line 16: if the directory is already created it causes an error
`FileExistsError: [Errno 17] File exists`

IvanRamyk (changes requested):

app/main.py
Terrrya: Done

app/main.py
```diff
@@ -1 +1,23 @@
...
+        print("Please enter the right command")
+        return
+    if "/" in command_list[2]:
+        path_list = command_list[2].split("/")
+    if "\\" in command_list[2]:
+        path_list = command_list[2].split("\\")
+    path = ""
+    for i in range(len(path_list) - 1):
+        path = os.path.join(path, path_list[i])
+        try:
```
IvanRamyk, line 16: it's better to check if dir is already present using `os.path.exists`

IvanRamyk (changes requested):

app/main.py
Terrrya: os.path.exists used

IvanRamyk (approved):
Good job!

PR #117, opened 2022-10-30
Solution

SpecializedBaby:


app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    # That will move a file from one location to another
```
IvanRamyk, line 5: it's better to use docstring to describe function behavior

app/main.py
```diff
@@ -1 +1,21 @@
...
+        print("Command mv not found")
+        return None
+    dir_path = ""
+    file_in = command.split(" ")[1]
+    file_out = command.split(" ")[2]
+    if "/" in file_out:
+        dir_ls = file_out.split("/")
+        for i in range(len(dir_ls) - 1):
+            dir_path += dir_ls[i] + "/"
+            os.mkdir(dir_path)
```
IvanRamyk, line 17: if the dir is already created an error is raised:
```FileExistsError: [Errno 17] File exists: ```

**Review by IvanRamyk — CHANGES_REQUESTED** [review]

File: app/main.py [hunk 3]
```diff
@@ -1 +1,29 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    """ That will move a file from one location to another """
+    # Ex: mv file.txt first_dir/second_dir/third_dir/file2.txt
+    if "mv" not in command:
+        print("Command mv not found")
```
vsmutok, line 8: wrong logic

what if my command looks?
```
cp file.txt mv_dir/second_dir/third_dir/file2.txt
```

app/main.py
```diff
@@ -1 +1,29 @@
...
+        return
+
+    dir_path = ""
+    file_in = command.split(" ")[1]
+    file_out = command.split(" ")[2]
```
vsmutok, line 13: DRY 

use `command.split()` and unpacking

app/main.py
```diff
@@ -1 +1,29 @@
...
+
+    if "/" in file_out:
+        for path in file_out.split("/")[:-1]:
+            dir_path += path + "/"
+            try:
+                os.mkdir(dir_path)
+            except OSError:
+                print("Error creating directory")
```
vsmutok, line 21: try to use `os.makedirs` and `os.path.dirname`

vsmutok (changes requested):

app/main.py
SpecializedBaby: yes I tried but Im using windows and need to replace but if this program run on mac will be error.

app/main.py
```diff
@@ -1 +1,33 @@
...
+def move_file(command: str) -> None:
+    """ That will move a file from one location to another """
+    # Ex: mv file.txt first_dir/second_dir/third_dir/file2.txt
+
+    (cmd,
+     file_in,
+     file_out) = (
+        command.split(" ")[0],
+        command.split(" ")[1],
+        command.split(" ")[2])
```
EdAlekseiev, line 13: Do not use split three times
```suggestion
    (cmd,
     file_in,
     file_out) = command.split()
     ```

File: app/main.py [hunk 7]
```diff
@@ -1 +1,33 @@
...
+
+    if cmd != "mv":
+        raise ValueError("Command mv not found")
+
+    if "/" in file_out:
+        dir_path = ""
+
+        for directory in os.path.dirname(file_out).split("/"):
```
EdAlekseiev, line 21: Use `os.path.split` instead of `split("/")`

app/main.py
```diff
@@ -1 +1,33 @@
...
+            dir_path = os.path.join(dir_path, directory)
+
+            if not os.path.exists(dir_path):
+                os.makedirs(dir_path)
```
EdAlekseiev, line 25: You can simplify this code with using os.makedirs(dir_path, exist_ok=True)
```suggestion
            os.makedirs(dir_path, exist_ok=True)
```

EdAlekseiev (changes requested):

app/main.py
SpecializedBaby: this will give me a string, but I need to split it into / to create each directory

EdAlekseiev: os.path.split returns tuple of 2 elements, where first element is directories. It's enought to create directories

app/main.py
```diff
@@ -1 +1,29 @@
...
+    if cmd != "mv":
+        raise ValueError("Command mv not found")
+
+    if "/" in file_out:
+        dir_path = ""
+        directories, filename = os.path.split(file_out)
+
+        for directory in directories.split("/"):
+            dir_path = os.path.join(dir_path, directory)
+            os.makedirs(dir_path, exist_ok=True)
```
EdAlekseiev, line 21: Still not fixed
```suggestion
        dir_path, file_name = directories.split("/")
        os.makedirs(dir_path, exist_ok=True)
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

SpecializedBaby:
What happened again this Error "Changes requasted"

PR #118, opened 2022-10-31
Move dile

bythewaters:


app/main.py
```diff
@@ -1 +1,27 @@
...
+
+    for index in range(len(mv_file_path)):
+        if ".txt" in mv_file_path[index]:
+            second_filename += mv_file_path[index]
+            mv_file_path.pop(index)
+
+    with open(first_file, "r"), open(second_filename, "w"):
+        path1 = ""
+        for path in mv_file_path:
+            path1 += path + "/"
```
IvanRamyk, line 21: use `os.path.join` to concatenate parts of the path

app/main.py
```diff
@@ -1 +1,27 @@
...
+            os.mkdir(path1)
```
IvanRamyk, line 22: if dir already exists it causes error

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #122, opened 2022-11-13
Solution

Ida83:


app/main.py
```diff
@@ -1 +1,18 @@
...
+    if "/" in source_data[2]:
+        makedirs(new_path)
+        with open(source_data[1], "w+") as file_in, \
+             open(path.join(new_path, paths[-1]), "w") as file_out:
+            save_data = file_in.read()
+            file_out.write(save_data)
+        remove(source_data[1])
+    else:
+        with open(source_data[1], "w+") as file_in:
+            pass
```
Abnormaltype, line 17: what a sense of this?

Abnormaltype (changes requested):
you should check the input command

app/main.py
Ida83: I can't rename file if it not exist

Ida83: Maybe task for renaming not clear for me

Abnormaltype: yes, but you can it more simply, don't use empty context managers

Ida83: just open(source_data[1], "w+") ?

Abnormaltype: try to find method for rename files, if this file doesn't exist, raise an error

Abnormaltype (approved):
Well done!

PR #132, opened 2022-11-26
Solution

Soobig666:


app/main.py
```diff
@@ -1 +1,23 @@
...
+    list_with_info = command.split(" ")
+    file_in_name = list_with_info[1]
+    dirs = list_with_info[2].split("/")
+
+    if list_with_info[0] != "mv":
+        return
+    try:
+        makedirs(f"{dirs[0]}/{dirs[1]}/{dirs[2]}")
+    except OSError:
+        pass
```
YuriiKindrat, line 14: If there will be only one/two dirs, will it work ?

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -4,12 +4,16 @@
 def move_file(command: str) -> None:
     list_with_info = command.split(" ")
```
Serhii-Leonenko, line 5: whitespace is the default separator for split, remove

app/main.py
```diff
@@ -4,12 +4,16 @@
...
     file_in_name = list_with_info[1]
-    dirs = list_with_info[2].split("/")
+    dirs_name = ""
 
     if list_with_info[0] != "mv":
         return
```
Serhii-Leonenko, line 10: this condition should be after that you split command

app/main.py
```diff
@@ -1 +1,27 @@
...
+    except OSError:
+        pass
+
+    with (
+        open(file_in_name, "r") as file_in,
+        open(list_with_info[-1], "w") as file_out
+    ):
+        file_out.write(file_in.read())
+        file_in.close()
+        file_out.close()
```
Serhii-Leonenko, line 26: you don't need close() if you use context manager

app/main.py
```diff
@@ -1 +1,27 @@
...
+
+
+def move_file(command: str) -> None:
+    list_with_info = command.split(" ")
+    file_in_name = list_with_info[1]
+    dirs_name = ""
+
+    if list_with_info[0] != "mv":
+        return
+    try:
```
Serhii-Leonenko, line 11: also, you should add a condition if you don't have dirs in the command just use `os.rename()`

app/main.py
```diff
@@ -1 +1,27 @@
...
+        for string in list_with_info[2].split("/"):
+            if "txt" in string:
+                break
+            dirs_name += string + "/"
```
Serhii-Leonenko, line 15: use `os.path.join()`

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from os import makedirs, remove, path
+
+
+def move_file(command: str) -> None:
+    command = command.split()
```
OleksandrPozden, line 5: you can unpack command into the variables

OleksandrPozden (changes requested):

app/main.py
Soobig666: what do you mean? Unpack command "mv" to another variables?

Soobig666: > you can unpack command into the variables
what do you mean? Unpack command "mv" to another variables?

Abnormaltype: command, file_name, ... = command.split()

Abnormaltype (changes requested):

Abnormaltype (approved):

PR #133, opened 2022-11-29
Solution

nailia-mamiedova:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    arguments = command.split()
```
OleksandrPozden, line 5:
```suggestion
    arg1, arg2, arg3 = command.split()
```
This is a better option instead of using  ```arguments[0]```
P.S. don't name variables as arg1. Name it as it is

OleksandrPozden (changes requested):

IvanRamyk (approved):
Good job!

PR #135, opened 2022-11-30
Solution

Zetamag:


app/main.py
```diff
@@ -1 +1,36 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    tokens = command.split(" ")
```
YuriiKindrat, line 5: no need to write split(" "), it breaks string by the " " by default

YuriiKindrat (changes requested):

OleksandrPozden (approved):
Well done!

PR #149, opened 2022-12-23
done

phaishuk:


app/main.py
```diff
@@ -1 +1,32 @@
...
+                pass
+        if len(path_list) > 1:
+            create_path(path_list)
+            with processing_files(
+                    f"{os.path.join(*path_list[:len(path_list)])}",
+                    "w",
+                    file_to_move,
+                    source
+            ):
+                pass
```
YuriiKindrat, line 32: no need to write pass here

YuriiKindrat (changes requested):

app/main.py
phaishuk: i've redone it at all, because it is possible to not only using os.path.join as it noticed in checklist. In chat i found that it is possible to use os.makedirs so i think it's much more efficient

OleksandrPozden (approved):

PR #151, opened 2022-12-24
Solution

evgenijmartynuk07:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    files = command.split()
```
IvanRamyk, line 5:
```
command_type, source, destination = command.split()
```
it's better to use unpacking here

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #153, opened 2022-12-26
solution

the-andrew-007:


app/main.py
```diff
@@ -1 +1,30 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    todo, file_from, file_to, *_ = command.split() + [None] * 2
```
olena-hul, line 5: For what you need `[None] * 2` ?

olena-hul (changes requested):

app/main.py
the-andrew-007: Oh, it should be [None] * 3
If split will return insufficient len list, `file_from`, `file_to` will be None
Thus user can pass an empty string or just mv in the function

app/main.py
```diff
@@ -1 +1,30 @@
...
+    dir_path_from = os.path.dirname(os.path.abspath(file_from))
+
+    if not os.path.exists(dir_path_to):
+        os.mkdir(dir_path_to)
+
+    if dir_path_from == dir_path_to:
+        os.rename(file_from, file_to)
+        return
+
+    with open(file_from, "r") as fs_file_from:
```
IvanRamyk, line 27: fs?

IvanRamyk: choose more clear names

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1 +1,30 @@
...
+    dir_path_from = os.path.dirname(os.path.abspath(file_from))
+
+    if not os.path.exists(dir_path_to):
+        os.mkdir(dir_path_to)
+
+    if dir_path_from == dir_path_to:
+        os.rename(file_from, file_to)
+        return
+
+    with open(file_from, "r") as file_stream_file_from:
```
Vasyl-Smutok, line 27: what does _file_stream_file_from_ variable mean? (файл_потік_файл_з)🧐
because it's hard for me to understand🥲

better to use this name  _stream_file_from_

app/main.py
```diff
@@ -1 +1,30 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    todo, file_from, file_to, *_ = command.split() + [None] * 3
+    if todo != "mv" or None in (file_from, file_to):
+        print("Usage: mv SOURCE DEST")
+        return
```
Vasyl-Smutok, line 8: why are you using empty `return`?
you can just delete it

app/main.py
```diff
@@ -1 +1,30 @@
...
+
+    if not os.path.exists(file_from):
+        print("File not found:", file_from)
+        return
```
Vasyl-Smutok, line 12: and here

app/main.py
```diff
@@ -1 +1,30 @@
...
+
+    dir_path_to = os.path.dirname(os.path.abspath(file_to))
+    dir_path_from = os.path.dirname(os.path.abspath(file_from))
+
+    if not os.path.exists(dir_path_to):
+        os.mkdir(dir_path_to)
+
+    if dir_path_from == dir_path_to:
+        os.rename(file_from, file_to)
+        return
```
Vasyl-Smutok, line 25: and here

app/main.py
```diff
@@ -1 +1,30 @@
...
+
+    with open(file_from, "r") as file_stream_file_from:
+        with open(file_to, "w") as file_stream_file_to:
```
Vasyl-Smutok, line 28: better to use this name stream_file_to

Vasyl-Smutok (changes requested):

app/main.py
the-andrew-007: Cos I want to interrupt further function execution, if user typed incorrect commandб further actions do not make sense

app/main.py
the-andrew-007: Same here, if user typed incorrect commandб further actions do not make sense

Abnormaltype (approved):
Well done!

PR #158, opened 2022-12-31
1st try solution (move-file)

AnnaKabatova:
I hope that it works at least, because I got lost couple times while solving this example 😅
And also there was no "pytests" so there were no guidelines also 😅

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,24 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_parts = command.split()
+    file_old = command_parts[1]
+    if "/" in command_parts[2]:
+        directories_list = command_parts[2].split("/")
+        file_new = directories_list[-1]
+        directories_list = directories_list[0:-1]
+        directories = "/".join(directories_list)
+        directories += "/"
+        os.makedirs(directories)
+        with open(file_old, "r") as file_in, open(file_new, "w") as file_out:
+            file_out.write(file_in.read())
+            file_in.close()
+            os.remove(f"{file_in}")
+            os.replace(f"{file_out}", f"{command_parts[2]}")
+    else:
+        file_new = command_parts[2]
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,24 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_parts = command.split()
+    file_old = command_parts[1]
```
OleksandrPozden, line 6:
```suggestion
    cmd, file_old, file_new = command.split()
```

app/main.py
```diff
@@ -1 +1,24 @@
...
+        directories_list = command_parts[2].split("/")
+        file_new = directories_list[-1]
+        directories_list = directories_list[0:-1]
+        directories = "/".join(directories_list)
+        directories += "/"
+        os.makedirs(directories)
+        with open(file_old, "r") as file_in, open(file_new, "w") as file_out:
+            file_out.write(file_in.read())
+            file_in.close()
+            os.remove(f"{file_in}")
```
OleksandrPozden, line 17: Don't repeat yourself

app/main.py
```diff
@@ -1 +1,24 @@
...
+
+def move_file(command: str) -> None:
+    command_parts = command.split()
+    file_old = command_parts[1]
+    if "/" in command_parts[2]:
+        directories_list = command_parts[2].split("/")
+        file_new = directories_list[-1]
+        directories_list = directories_list[0:-1]
+        directories = "/".join(directories_list)
+        directories += "/"
```
OleksandrPozden, line 12:
```suggestion
        directories = "/".join(directories_list)+"/"
```

OleksandrPozden (changes requested):

AnnaKabatova:
2nd commit should work, but it still isnt' working properly, so i need some help to understand why

YuriiKindrat (approved):

PR #163, opened 2023-01-08
Solution

yuliia-stopkyna:


katevasilenko (changes requested):
Please read checklist and make changes in your code

app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    short_command = command.split()[0]
+    source = command.split()[1]
+    path = command.split()[2]
```
olena-hul, line 7: You can use unpacking

app/main.py
```diff
@@ -1 +1,22 @@
...
+        return
+    if "/" not in path:
+        os.rename(source, path)
+        return
+    path_parts = path.split("/")[:-1]
+    filename = path.split("/")[-1]
+    path = ""
+    for part in path_parts:
+        path = os.path.join(path, part)
+    os.makedirs(path, exist_ok=True)
```
olena-hul, line 18: You can use this to obtain a directory name `destination_directory = os.path.dirname(path)`. And then you can just make `os.makedirs(destination_directory, exist_ok=True)`

app/main.py
```diff
@@ -1 +1,22 @@
...
+    filepath = os.path.join(path, filename)
+    with open(filepath, "w") as new_file, open(source, "r") as copy_from:
```
olena-hul, line 20: `copy_from` is not a super desctriptive name. `source_file` would be better

olena-hul (changes requested):

katevasilenko (approved):
Looks good to me

PR #169, opened 2023-01-18
Solution

YehorKorn:


app/main.py
```diff
@@ -1 +1,46 @@
...
+        for current_dir in path_catalog[:-1]:
+            creating_dir.append(current_dir)
+            if not os.path.isdir("/".join(creating_dir)):
+                os.mkdir("/".join(creating_dir))
+
+        if os.path.isdir("/".join(path_catalog[:-1])):
+            move_clear_file(command, file_path_name1, file_path_name2)
+
+    if len(path.split("/")) == 1:
+        move_clear_file(command, file_path_name1, file_path_name2)
```
ArsenPidhoretskyi, line 46: Can you explain why you need this? Why not just do move_clear_file(command, file_path_name1, file_path_name2)

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,37 @@
...
+    ):
+        file_out.write(file_in.read())
+        file_in.close()
+        os.remove(file_path1)
+
+
+def move_file(command: str) -> None:
+    if len(command.split()) != 3:
+        raise ValueError("This is not valid")
+    command, file_path_name1, file_path_name2 = command.split()
```
olena-hul, line 18:
```suggestion
    command, source_file, target_file = command.split()
```

app/main.py
```diff
@@ -1 +1,37 @@
-# write your code here
+from __future__ import annotations
+import os
+
+
+def move_clear_file(file_path1: str, file_path2: str) -> None:
```
olena-hul, line 5:
```suggestion
def move_clear_file(source_file: str, target_file: str) -> None:
```

app/main.py
```diff
@@ -1 +1,37 @@
...
+    with (
+        open(file_path1, "r") as file_in,
+        open(file_path2, "w") as file_out
+    ):
+        file_out.write(file_in.read())
+        file_in.close()
```
olena-hul, line 11: No need to close file manually, the close will be done after with block is ended

app/main.py
```diff
@@ -1 +1,37 @@
...
+    if len(command.split()) != 3:
+        raise ValueError("This is not valid")
+    command, file_path_name1, file_path_name2 = command.split()
+    if command != "mv":
+        raise ValueError(
+            f'"{command}" This is an invalid command!'
+            f' The correct one is: "mv"'
+        )
+
+    path = os.path.join(file_path_name2)
```
olena-hul, line 25: Why you need to call this if there is only 1 path to be joined?

app/main.py
```diff
@@ -1 +1,37 @@
...
+
+    if not head_path and tail_path:
+        move_clear_file(file_path_name1, file_path_name2)
+
+    if head_path and tail_path:
+        current_path = ""
+        for current_dir in head_path.split("/"):
+            current_path = os.path.join(current_path, current_dir)
+            if not os.path.isdir(current_path):
+                os.mkdir(current_path)
```
olena-hul, line 36: You can use makedirs function from os module, to avoid creating directories in a loop

olena-hul (changes requested):

app/main.py
YehorKorn: To delete a file inside the context manager you have to close it manually, otherwise it won't let you delete it

app/main.py
YehorKorn: In order for the path to be correct for all OS.

app/main.py
olena-hul: This comment is not fixed

app/main.py
```diff
@@ -1 +1,37 @@
...
+
+    if not head_path and tail_path:
+        move_clear_file(source_file, target_file)
+
+    if head_path and tail_path:
+        current_path = ""
+        for current_dir in head_path.split("/"):
+            current_path = os.path.join(current_path, current_dir)
+            if not os.path.isdir(current_path):
+                os.mkdir(current_path)
```
olena-hul, line 36: The comment about using makedirs is not fixed

app/main.py
olena-hul: Okay, let it be

olena-hul (changes requested):

app/main.py
ArsenPidhoretskyi: @YehorKorn maybe then delete the file after context manager?

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+    path = os.path.join(target_file)
+    head_path, tail_path = os.path.split(path)
+
+    if not head_path and tail_path:
+        move_clear_file(source_file, target_file)
+
+    if head_path and tail_path:
+        os.makedirs(head_path)
+        move_clear_file(source_file, target_file)
```
ArsenPidhoretskyi, line 33: 33 and 29 line duplicates
Maybe check for head_path then os.makedirs
And then check for the tail_path and the move_clear_file

app/main.py
```diff
@@ -1 +1,33 @@
...
+    if len(command.split()) != 3:
+        raise ValueError("This is not valid")
+    command, source_file, target_file = command.split()
+    if command != "mv":
+        raise ValueError(
+            f'"{command}" This is an invalid command!'
+            f' The correct one is: "mv"'
+        )
+
+    path = os.path.join(target_file)
```
ArsenPidhoretskyi, line 25: why do you need os.path.join with one argument?

ArsenPidhoretskyi (changes requested):

app/main.py
ArsenPidhoretskyi: @YehorKorn check this

ArsenPidhoretskyi (changes requested):
Check comment

Vasyl-Smutok (approved):
Well Done 👍

PR #171, opened 2023-01-20
Solution

alina-boichenko:


Vasyl-Smutok (changes requested):
what if you have multiple folders?
image

app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if len(command.split()) == 3:
```
olena-hul, line 5: Move command.split() to a variable

app/main.py
```diff
@@ -1 +1,21 @@
...
+        command_name, file_name, destination_path = command.split()
+
+        if "/" in destination_path:
+            destination_path = destination_path.split("/")
+            new_file_name = destination_path[-1]
+            destination_path = destination_path[:-1]
+            for path in destination_path:
+                os.makedirs(path)
```
olena-hul, line 13: makedirs should be used to create directories without using loop, you can just pass the full path (exist_ok=True argument should be passed additionally to avoid os errors if the directory already exists)

olena-hul (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #173, opened 2023-01-21
'Solution'

timyr4iq:


app/main.py
```diff
@@ -1 +1,17 @@
...
+
+def move_file(command: str) -> None:
+    files_info = command[3:].split(" ")
+    first_file = files_info[0]
+    second_file_info = files_info[1].split("/")
+    if len(second_file_info) == 1:
+        os.rename(first_file, second_file_info[0])
+    else:
+        path = f"{second_file_info[0]}"
+        for i in range(1, len(second_file_info)):
```
Vasyl-Smutok, line 12: `i` bad variable name

Vasyl-Smutok (changes requested):
image

app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    try:
+        mv, current_file, copied_file_path = command.split()
```
ArsenPidhoretskyi, line 6: 1. mv is not the correct naming, it is 'command'
2. You need to validate the command. What happens if there are more than 3 values?

app/main.py
```diff
@@ -1 +1,21 @@
...
+    except ValueError:
+        raise ValueError("Command is incorrect")
+    if mv != "mv" or os.path.exists(current_file) is False:
```
ArsenPidhoretskyi, line 9:
```suggestion
    if mv != "mv" or not os.path.exists(current_file):
```

app/main.py
```diff
@@ -1 +1,21 @@
...
+        raise ValueError("Command is incorrect")
+    copied_file_info = copied_file_path.split("/")
+    if len(copied_file_info) == 1:
+        os.rename(current_file, copied_file_info[0])
+    else:
+        path = f"{copied_file_info[0]}"
```
ArsenPidhoretskyi, line 15: No need f-string here

app/main.py
```diff
@@ -1 +1,21 @@
...
+        for index_of_folder in range(1, len(copied_file_info)):
+            os.mkdir(path)
+            path += f"/{copied_file_info[index_of_folder]}"
```
ArsenPidhoretskyi, line 18: Better to use os.path.join, not add string

ArsenPidhoretskyi (changes requested):
Also, if you want to do with `os.rename(current_file, copied_file_info[0])`
You can create all folders and just do `os.rename(current_file, copied_file_path)`

app/main.py
```diff
@@ -1 +1,17 @@
...
+    except ValueError:
+        raise ValueError("Command is incorrect")
+    if input_command != "mv" or not os.path.exists(current_file):
+        raise ValueError("Command is incorrect")
+    copied_file_info = copied_file_path.split("/")
+    if len(copied_file_info) > 1:
+        path = ""
+        for folder_name in copied_file_info[:-1]:
+            path = os.path.join(path, folder_name)
+            os.mkdir(path)
```
olena-hul, line 16: you can use `os.makedirs(path)` to create all directories at once, not using loop

olena-hul (changes requested):

PR #176, opened 2023-01-26
Solution

TarasSavchyn:


app/main.py
```diff
@@ -1 +1,22 @@
...
+        if "/" not in new_file:
+            with open(new_file, "w") as new:
+                new.write(context)
+        else:
+            directories = new_file.split("/")[:-1]
+            new_file = new_file.split("/")[-1]
+            for directory in directories:
+                path = os.path.join(path, directory)
+                os.mkdir(path)
+            with open(f"{path}/{new_file}", "w") as file:
```
Bohdan-Salamakha, line 20: use  os.path.join here too, it's a good practice cause u might work on linux where path used backslashes

app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    old_file, new_file = command.split()[1:]
+    to_del = old_file
```
Bohdan-Salamakha, line 6: u don't need this temp variable (to_del), just use old_file when u wanna remove old file

app/main.py
```diff
@@ -1 +1,22 @@
...
+    path = ""
+
+    with open(old_file, "r") as old:
```
Bohdan-Salamakha, line 9: not `as old`, better is `as file_read_stream` or just `file_read`

Bohdan-Salamakha (changes requested):
v1

app/main.py
```diff
@@ -1 +1,23 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if command.split()[0] == "mv":
```
ArsenPidhoretskyi, line 5: You need to validate command
In this line I already get an error (command = "")
Check your code for this cases:
mv file1.txt
mv file1.txt file2.txt file3.txt

app/main.py
```diff
@@ -1 +1,23 @@
...
+        file_to_read, file_to_write = command.split()[1:]
+        old_file = file_to_read
```
ArsenPidhoretskyi, line 7: Why you need this?

app/main.py
```diff
@@ -1 +1,23 @@
...
+                    file_write.write(context)
+            else:
+                directories = file_to_write.split("/")[:-1]
+                new_file = file_to_write.split("/")[-1]
+                for directory in directories:
+                    path = os.path.join(path, directory)
+                    os.mkdir(path)
+                with open(os.path.join(path, new_file), "w") as file_write:
+                    file_write.write(context)
+            os.remove(old_file)
```
ArsenPidhoretskyi, line 23: You trying to remove opened file/ Remove it after context manager

app/main.py
```diff
@@ -1 +1,23 @@
...
+            context = file_read.read()
+            if "/" not in file_to_write:
+                with open(file_to_write, "w") as file_write:
+                    file_write.write(context)
+            else:
+                directories = file_to_write.split("/")[:-1]
+                new_file = file_to_write.split("/")[-1]
+                for directory in directories:
+                    path = os.path.join(path, directory)
+                    os.mkdir(path)
```
ArsenPidhoretskyi, line 20: You can use os.makedirs instead of os.makedir
Also, os.path.join can work as: os.path.join(start_path, *[list values])

app/main.py
```diff
@@ -1 +1,23 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if command.split()[0] == "mv":
+        file_to_read, file_to_write = command.split()[1:]
+        old_file = file_to_read
+        path = ""
```
ArsenPidhoretskyi, line 8: It is not good to declare a variable so early, you use it only in else

app/main.py
```diff
@@ -1 +1,23 @@
...
+                with open(file_to_write, "w") as file_write:
+                    file_write.write(context)
+            else:
+                directories = file_to_write.split("/")[:-1]
+                new_file = file_to_write.split("/")[-1]
+                for directory in directories:
+                    path = os.path.join(path, directory)
+                    os.mkdir(path)
+                with open(os.path.join(path, new_file), "w") as file_write:
+                    file_write.write(context)
```
ArsenPidhoretskyi, line 22: Almost the same code as 13-14 line

app/main.py
```diff
@@ -1 +1,23 @@
...
+import os
+
+
+def move_file(command: str) -> None:
+    if command.split()[0] == "mv":
+        file_to_read, file_to_write = command.split()[1:]
+        old_file = file_to_read
+        path = ""
+
+        with open(file_to_read, "r") as file_read:
```
ArsenPidhoretskyi, line 10: Maybe better first create all dirs and then just read-write-delete file?

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if command.split()[0] == "mv" and len(command.split()) == 3:
+        file_to_read, file_to_write = command.split()[1:]
+        with open(file_to_read, "r") as file_read:
+            context = file_read.read()
```
Bohdan-Salamakha, line 8: Не тримай read stream активним, ти ж вже записав context у RAM
перейди на 1 левел вложеності ліворуч після запису контекста

Bohdan-Salamakha (changes requested):
v1

PR #177, opened 2023-01-28
Solution

olenaliuby:


app/main.py
```diff
@@ -1 +1,44 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    try:
+        mv_command, source_file, dstn_path = command.split()
```
Vasyl-Smutok, line 6: `dstn_path` bad variable name 🥲

Vasyl-Smutok (changes requested):

app/main.py
olenaliuby: Thank you for highlighting this!
Changed `dstn_path` to `destination`. It will be better?

app/main.py
```diff
@@ -1 +1,44 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    try:
+        mv_command, source_file, destination = command.split()
+    except ValueError:
+        raise ValueError(
```
katevasilenko, line 8: It’s a bad practice to catch exception and then raise this again. You can just print message, it’s  a good catch

katevasilenko (changes requested):

app/main.py
olenaliuby: Thank you! Just changed it. Is other ValueError raising is fine?

Abnormaltype (changes requested):
Good job!
But your function has many things.
Split it into several functions.

app/main.py
```diff
@@ -1 +1,52 @@
...
+            f"Do you want rewrite {destination} with {source_file}: yes/no?"
+        )
+        if user_input.lower() == "no":
+            raise ValueError("Please, rename destination file")
+
+
+def move_file(command: str) -> None:
+    try:
+        mv_command, source_file, destination = command.split()
+    except ValueError:
```
ArsenPidhoretskyi, line 37: except ValueError and print() than fail.
 Use raise ValueError(your message) instead of print

app/main.py
```diff
@@ -1 +1,52 @@
...
+
+def create_dirs_from_str(destination: str) -> None:
+    if "/" in destination:
+        path = os.path.split(destination)[0]
+        if not os.path.exists(path):
+            os.makedirs(path)
+
+
+def relocate_file_if_path_exists(source_file: str, destination: str) -> None:
+    if os.path.isdir(destination):
```
ArsenPidhoretskyi, line 17: if os.path.isdir(destination):
It is not good. the destination is dir than how you can move file?

ArsenPidhoretskyi: You solution is okay, but in the task we do not have this case

app/main.py
```diff
@@ -1 +1,52 @@
...
+def check_if_mv(command: str) -> None:
+    if command != "mv":
+        raise ValueError("Command must begging with 'mv' without flags")
+
+
+def create_dirs_from_str(destination: str) -> None:
+    if "/" in destination:
+        path = os.path.split(destination)[0]
+        if not os.path.exists(path):
+            os.makedirs(path)
```
ArsenPidhoretskyi, line 13: os.makedirs(path, exist_ok=True)

app/main.py
```diff
@@ -1 +1,52 @@
...
+    rewrite_existed_file(source_file, destination)
+
+    with (
+        open(source_file, "r") as source,
+        open(destination, "w") as output_file
+    ):
+        output_file.write(source.read())
+
+    if destination != source_file:
+        os.remove(source_file)
```
ArsenPidhoretskyi, line 52: Do it before write, just check if not the same files

app/main.py
```diff
@@ -1 +1,52 @@
...
+        mv_command, source_file, destination = command.split()
+    except ValueError:
+        print("Command must have format: mv source_file destination/file_name")
+
+    check_if_mv(mv_command)
+    relocate_file_if_path_exists(source_file, destination)
+    create_dirs_from_str(destination)
+    rewrite_existed_file(source_file, destination)
+
+    with (
```
ArsenPidhoretskyi, line 45: It is strange:
1. destination is dir but you trying to write

ArsenPidhoretskyi (changes requested):

app/main.py
olenaliuby: But [a bit earlier, Kate wrote](https://github.com/mate-academy/py-move-file/pull/177#discussion_r1092419721), that's bad practice.
Which way is better?

app/main.py
olenaliuby: I just thought about different possible cases. Can I leave it, or better to remove it?

app/main.py
olenaliuby: In this case I work with possibility, when in destination is file, that already exist.
So I ask user if they wanna rewrite an existing file. In task there are no such cases, but in life it's possibility.

`def rewrite_existed_file(source_file: str, destination: str) -> None:
    if os.path.isfile(destination):
        user_input = input(
            f"Do you want rewrite {destination} with {source_file}: yes/no?"
        )
        if user_input.lower() == "no":
            raise ValueError("Please, rename destination file")`

Vasyl-Smutok (approved):
Well Done 👍

PR #180, opened 2023-02-01
Solution

ilushasm:


README.md
```diff
@@ -18,8 +18,8 @@ create file2.txt inside and remove the source file.
 print(open("file.txt").read())
 # Some
 # Text
-move_file("mv file.txt first_dir/second_dir/third_dir/file2.txt")
-print(open("first_dir/second_dir/third_dir/file2.txt").read())
+move_file("mv file.txt first_dir/second_dir/third_dir/file.txt")
+print(open("first_dir/second_dir/third_dir/file.txt").read())
```
katevasilenko, line 22: Please don’t fix readme file

app/main.py
```diff
@@ -1 +1,30 @@
-# write your code here
+from pathlib import Path
+import os
+
+
+def move_file(command: str) -> None:
+    comm_list = command.split(" ")
```
katevasilenko, line 6: You can declare all three variables here in one line something like (also don’t use space in split as it use it by default)
```suggestion
    command, old_file, new_file = command.split()
```

app/main.py
```diff
@@ -1 +1,30 @@
...
+
+    if comm_list[0] == "mv":
+        created_dir = ""
+        new_file_path = list(Path(comm_list[2]).parts)
+        orig_file = comm_list[1]
+        n_f_name = orig_file
```
katevasilenko, line 12: Bad naming here

app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+from pathlib import Path
+import os
+
+
+def move_file(command: str) -> None:
+    command, old_file, new_file = command.split()
```
olena-hul, line 6: Validate length of command too, it should be exactly 3 items

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+        if "." in new_file_path[-1].strip():
+            new_file_name = new_file_path[-1]
+            new_file_path.pop()
+
+        if len(new_file_path) > 1:
+            for index in range(len(new_file_path)):
+                created_dir = os.path.join(created_dir, new_file_path[index])
+                if not os.path.isdir(created_dir):
+                    os.mkdir(created_dir)
```
olena-hul, line 21: You can just retrieve dir path with `os.path.dirname` or `os.path.split` and use `os.makedirs` to create directory at once without using a loop

olena-hul (changes requested):

app/main.py
ilushasm: and how to validate that string has 3 words without splitting it?

app/main.py
```diff
@@ -1 +1,28 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if len(command.split()) == 3:
```
ArsenPidhoretskyi, line 5: Better to do 
```
if len(commnad.split()) != 3:
    and the raise 
```
This is named as golden path

app/main.py
```diff
@@ -1 +1,28 @@
...
+        command, old_file, new_file = command.split()
+
+        if command == "mv":
+            new_file_path, new_file_name = os.path.split(new_file)
+
+            if len(new_file_path) > 0:
+                if not os.path.isdir(new_file_path):
+                    os.makedirs(new_file_path)
```
ArsenPidhoretskyi, line 13: os.makedirs(new_file_path, exist_ok=True)

app/main.py
```diff
@@ -1 +1,28 @@
...
+
+            if len(new_file_name) == 0:
+                new_file_name = old_file
+
+            print(new_file_path + new_file_name)
+
+            with open(old_file, "r") as file_old, open(
+                    os.path.join(new_file_path, new_file_name),
+                    "w"
+            ) as file_copy:
```
ArsenPidhoretskyi, line 23:
```suggestion
            destination_path = os.path.join(new_file_path, new_file_name)
            with (
                open(old_file, "r") as file_old, 
                open(destination_path, "w") as file_copy:
            ):
```

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,27 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if len(command.split()) != 3:
+        raise
```
Grayder0152, line 6: You must raise some exception, for example ValueError

Grayder0152 (changes requested):

ArsenPidhoretskyi (approved):
Looks good for me

PR #189, opened 2023-02-13
Solution

manjustice:


AndereLion (approved):

app/main.py
```diff
@@ -1 +1,19 @@
...
+
+
+def move_file(command: str) -> None:
+    command, file1, file2_path = command.split()
+    *dirs, file2 = file2_path.split("/")
+
+    if command == "mv":
+        if not dirs:
+            rename(file1, file2)
+            return
```
YuriiKindrat, line 11: no need to use blank return use else instead

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #190, opened 2023-02-13
tests added

Y-Havryliv:


IvanRamyk (changes requested):

tests/test_main.py
```diff
@@ -0,0 +1,62 @@
...
+    assert os.path.exists("file.txt") is False
+
+    with open("first_dir/second_dir/file2.txt", "r") as file_with_content:
+        assert file_with_content.read() == "This is some\n content for\n the file."
+
+    shutil.rmtree("first_dir")
+
+
+def test_should_create_multiple_directories_when_they_exist(create_file: callable) -> None:
+    os.makedirs("first_dir/second_dir/third_dir")
```
IvanRamyk, line 54:
```suggestion
    os.makedirs("first_dir/second_dir")
```
I meant that only part of the path should exist

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #191, opened 2023-02-13
1 commit

AndereLion:


app/main.py
```diff
@@ -1 +1,22 @@
...
+
+def move_file(command: str) -> None:
+    command_mv, file_to_opening, directory_to_writing = command.split(" ")
+    if command_mv == "mv":
+        if directory_to_writing.count("/"):
+
+            directory_to_writing_in_list = directory_to_writing.split("/")
+            file_to_writing = directory_to_writing_in_list[-1]
+            folders = ""
+            for ind in range(len(directory_to_writing_in_list) - 1):
```
YuriiKindrat, line 13: ind is a bad name for variable, use full names for naming
also u can use makedirs method instead of loop)

YuriiKindrat (changes requested):

app/main.py
AndereLion: Fixed

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+import os
+import shutil
+
+
+def move_file(command: str) -> None:
+    command, file_to_opening, path = command.split()
+    if command == "mv":
+        if path.count("/"):
+            path_to_writing = path.split("/")
```
Serhii-Leonenko, line 9: Make your code universal, use os.path.split(). For example, windows use `\` instead of `/`

app/main.py
```diff
@@ -1 +1,15 @@
...
+            os.makedirs(os.path.join(*path_to_writing[0:-1]))
+            shutil.copyfile(file_to_opening, os.path.join(*path_to_writing))
+            os.remove(file_to_opening)
+        else:
+            shutil.copyfile(file_to_opening, path)
+            os.remove(file_to_opening)
```
Serhii-Leonenko, line 15: `You can use os.rename() instead of this`

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+import os
+import shutil
+
+
+def move_file(command: str) -> None:
+    command, file_to_opening, path = command.split()
```
Serhii-Leonenko, line 6: path is not a descriptive name in this case

Serhii-Leonenko (changes requested):

app/main.py
AndereLion: Fixed

app/main.py
AndereLion: Fixed

app/main.py
AndereLion: Done

app/main.py
```diff
@@ -1 +1,15 @@
...
+import shutil
+
+
+def move_file(command: str) -> None:
+    command, file_to_opening, path_to_file = command.split()
+
+    if command == "mv":
+        if path_to_file.count("/"):
+            path_to_writing = os.path.split(path_to_file)
+            os.makedirs(os.path.join(*path_to_writing[0:-1]))
```
Serhii-Leonenko, line 11: can you make directories directly with the head from os.path.split ?)

Serhii-Leonenko (changes requested):

app/main.py
AndereLion: Done

YuriiKindrat (approved):

PR #192, opened 2023-02-13
Solution

SafonovVladimir:
add my solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,16 @@
-# write your code here
+import os
+
+
+def move_file(command_line: str) -> None:
+    command, first_file, second_file = command_line.split(" ")
+    last_slash_index = second_file.rfind("/")
+    second_file_path = second_file[:last_slash_index]
+    content = ""
+    os.makedirs(second_file_path)
+
+    with open(first_file, "r") as file_content:
+        content += file_content.read()
+
+    with open(second_file, "w") as file_out:
+        file_out.write(content)
+        os.remove(first_file)


MaybeTI (changes requested):
You should check which command we need to execute
if command == "mv"

app/main.py
```diff
@@ -1 +1,20 @@
...
+        if not os.path.exists(second_file_path):
+            os.makedirs(second_file_path)
+
+        with open(first_file, "r") as file_content:
+            content += file_content.read()
+
+        os.remove(first_file)
+
+        with open(second_file, "w") as file_out:
+            file_out.write(content)
```
YuriiKindrat, line 20: no need to use CM for 2 times, try to make it using with only once

app/main.py
```diff
@@ -1 +1,20 @@
...
+import os
+
+
+def move_file(command_line: str) -> None:
+    command, first_file, second_file = command_line.split()
+    last_slash_index = second_file.rfind("/")
+    second_file_path = second_file[:last_slash_index]
+    content = ""
+
+    if command == "mv" and last_slash_index != "-1":
```
YuriiKindrat, line 10: are u sure about "-1" condition ?

YuriiKindrat (changes requested):

app/main.py
SafonovVladimir: Of couse no, we can use just one.
Fixed

app/main.py
SafonovVladimir: Check it, we don't need it.
Delete

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command, first_file, second_file = command.split()
+    second_file_path = second_file[:second_file.rfind("/")]
```
Serhii-Leonenko, line 6: it's better to use `os.path.split()` for this purpose, for example windows use `\` instead of `/`

app/main.py
```diff
@@ -1 +1,19 @@
...
+            os.makedirs(second_file_path)
+
+        with (open(first_file, "r") as file_content,
+              open(second_file, "w") as file_out):
+            file_out.write(file_content.read())
+
+        os.remove(first_file)
+
+
+move_file("mv text.txt dir1/dir2/dir3/text.txt")
```
Serhii-Leonenko, line 19: it's redundant here

Serhii-Leonenko (changes requested):

app/main.py
SafonovVladimir: Forgot to delete it
Done

app/main.py
Serhii-Leonenko: what about it?)

Serhii-Leonenko (changes requested):

app/main.py
SafonovVladimir: I rewrite this part of code
Check please

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command, first_file, second_file = command.split()
+    second_file_part = os.path.split(second_file)
+    directories_list = second_file_part[0].split("/")
```
Serhii-Leonenko, line 7: do you need this `directories_list = second_file_part[0].split("/")` ? Can you create directories without it?)

Serhii-Leonenko (changes requested):

app/main.py
SafonovVladimir: U r right. I don't need several lines. Thanks

kuznetsb (approved):
Seems okay

README.md
```diff
@@ -15,15 +15,15 @@ Examples:
...
+print(open("first_dir/second_dir/third_dir/text.txt").read())
 # Some
 # Text
-open("file.txt")
-# FileNotFoundError: [Errno 2] No such file or directory: 'file.txt'
+open("text.txt")
+# FileNotFoundError: [Errno 2] No such file or directory: 'text.txt'
 ```
 
 **Note**: You can create directory by `os.mkdir("first")` from module `os`.
```
YuriiKindrat, line 29: no need to change readme.md file)

YuriiKindrat (approved):

PR #193, opened 2023-02-13
Solution

MaybeTI:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    func, file, directory = command.split()
```
bihunva, line 5: comment

*checklist

bihunva (changes requested):

NEkropo1 (changes requested):
image
image
Refactor according to checklist, please )

MaybeTI:
> image
> image
> Refactor according to checklist, please )

@NEkropo1 
I really don't know how to use this module, I've already tried everything

NEkropo1:
> 

you should do this command with path, like if u have os.path.join("path_dir1/path_dir2/") it will reorganize it according to your system path methods.

app/main.py
```diff
@@ -1 +1,17 @@
...
+    if func == "mv":
+        if "/" in command:
+            command_ls = directory.split("/")
+            for direc in command_ls[:-1]:
```
YuriiKindrat, line 9: direc is bad name for variable use full names for naming

app/main.py
```diff
@@ -1 +1,17 @@
...
+                os.mkdir(f"{directory[: directory.index(direc)]}{direc}")
```
YuriiKindrat, line 10: try to simplify ur code here, it is realy hard to read it

app/main.py
```diff
@@ -1 +1,17 @@
...
+
+            with open(file, "r") as first_file, open(directory, "w") as copy:
+                copy.write(first_file.read())
+
+            os.remove(file)
+        else:
+            os.rename(file, directory)
```
YuriiKindrat, line 17: it should work without this part of code

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,18 @@
...
+            for direction in command_ls[:-1]:
+                file_directory = directory[: directory.index(direction)]
+                os.mkdir(f"{file_directory}{direction}")
+
+            with open(file, "r") as first_file, open(directory, "w") as copy:
+                copy.write(first_file.read())
+
+            os.remove(file)
+        else:
+            os.rename(file, directory)
```
YuriiKindrat, line 18: will 13-16 lines do the same thing as rename, if there is won't be "/" in command ? No need to use else, here, try to reuse 13-26 lines

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,15 @@
...
+import os
+
+
+def move_file(command: str) -> None:
+    command, file, directory = command.split()
+    if command == "mv":
+        command_ls = directory.split("/")
+        for direction in command_ls[:-1]:
+            file_directory = directory[: directory.index(direction)]
+            os.mkdir(f"{file_directory}{direction}")
```
YuriiKindrat, line 10: make a check for " / " for this part of code, it won't work correctly with command without " / "

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,18 @@
...
+def move_file(command: str) -> None:
+    command, file, directory = command.split()
+    if command == "mv":
+        if "/" in directory:
+            command_ls = directory.split("/")
+        else:
+            command_ls = directory
+        for direction in command_ls[:-1]:
+            file_directory = directory[: directory.index(direction)]
+            os.mkdir(f"{file_directory}{direction}")
```
Serhii-Leonenko, line 13: Is your code cross-platform?)For example, will it work with \ in command on windows?)

HINT: you can try to use os.path.split() to split directories and file_name in the path, os.makedirs() - to create directories without loop, and whatever you want)

Serhii-Leonenko (changes requested):

PR #194, opened 2023-02-14
'Solution'

serhii-skivka:


app/main.py
```diff
@@ -1 +1,19 @@
...
+
+
+def move_file(command: str) -> None:
+    mv, file, dir_file = command.split()
+    dir_file_split = dir_file.split("/")
+
+    if mv != "mv":
+        return
+    if file == dir_file:
+        return
```
Bohdan-Salamakha, line 12:
```
    if mv != "mv" or file == dir_file:
        return
```

app/main.py
```diff
@@ -1 +1,19 @@
...
+    new_path = ""
+    for path in dir_file_split[:-1]:
+        new_path = os.path.join(new_path, path)
+    os.mkdir(new_path)
+    with open(file, "r") as file_in, open(dir_file_split[-1]) as file_out:
+        file_out.write(file_in.read())
+        os.remove(file)
```
Bohdan-Salamakha, line 19: place this line on one tab level to left
`os.remove(file)`

```
    with open(file, "r") as file_in, open(dir_file_split[-1]) as file_out:
        file_out.write(file_in.read())
    os.remove(file)
```

Bohdan-Salamakha (changes requested):
v1

app/main.py
```diff
@@ -1 +1,19 @@
 # write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    mv, file, dir_file = command.split()
```
ArsenPidhoretskyi, line 6: file - is the bad name it this case, first of all, it is source_file_path
dir_file - also not a good name. I totally does not understand what it is mean) Use destination_file_path
Also validate command for 3 params
mv - it is not mv it input_command

app/main.py
ArsenPidhoretskyi, line 12: Write inline this by using or

app/main.py
```diff
@@ -1 +1,19 @@
...
+    dir_file_split = dir_file.split("/")
+
+    if mv != "mv":
+        return
+    if file == dir_file:
+        return
+    new_path = ""
+    for path in dir_file_split[:-1]:
+        new_path = os.path.join(new_path, path)
```
ArsenPidhoretskyi, line 15: Why you need this?
Use: os.path.join(*dir_file_split)

app/main.py
```diff
@@ -1 +1,19 @@
...
+    os.mkdir(new_path)
```
ArsenPidhoretskyi, line 16: Use os.makedirs with exist_ok=True. it is also possible to not only one dir

app/main.py
```diff
@@ -1 +1,19 @@
 # write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    mv, file, dir_file = command.split()
+    dir_file_split = dir_file.split("/")
```

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,19 @@
 # write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    input_command, source_file_path, destination_file_path = command.split()
```
olena-hul, line 6: Move command.split() to a separate variable because you're duplicating it in line 11

app/main.py
```diff
@@ -1 +1,19 @@
...
+
+    if input_command != "mv":
+        raise NameError("You`ve entered wrong command")
+
+    if input_command == "mv" and len(command.split()) == 3:
+        if "/" in destination_file_path:
+            os.makedirs(os.path.dirname(destination_file_path))
+            with open(f"{source_file_path}", "r") as file_in, \
+                    open(f"{destination_file_path}", "w") as file_out:
```
olena-hul, line 15:
```suggestion
            with (
                open(f"{source_file_path}", "r") as file_in,
                open(f"{destination_file_path}", "w") as file_out
            ):
```

olena-hul (changes requested):

app/main.py
```diff
@@ -1 +1,22 @@
...
+    if input_command == "mv" and len(command) == 3:
+        if "/" in destination_file_path:
+            os.makedirs(os.path.dirname(destination_file_path))
+
+            with (
+                open(f"{source_file_path}", "r") as file_in,
+                open(f"{destination_file_path}", "w") as file_out
+            ):
+                file_out.write(file_in.read())
+                os.remove(source_file_path)
```
Bohdan-Salamakha, line 20: place it one level tabulation to the left cause u don't need the IO stream to be opened to remove file

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #198, opened 2023-02-16
Learn flow

NEkropo1:


ZabFTFT (approved):

app/main.py
```diff
@@ -1 +1,28 @@
...
+
+
+def move_file(exec_string: str) -> None:
+    commands = {"mv"}
+    command, old_file, new_file = exec_string.split()
+    if command not in commands:
+        raise AttributeError(f"Allowed commands: {commands},"
+                             f" {command} is not allowed!")
+    path1, path2 = os.path.split(old_file)[0], os.path.split(new_file)[0]
+    path1, path2 = os.path.join(path1), os.path.join(path2)
```
YuriiKindrat, line 15: why do u need this part ?

app/main.py
```diff
@@ -1 +1,28 @@
...
+
+    if path1 == path2:
+        if old_file == new_file:
+            return
```
YuriiKindrat, line 19: try to avoid using blank return

YuriiKindrat (changes requested):

app/main.py
NEkropo1: since we can get any input with any type of os.path separators, these lanes split lanes for any mistakes in them and rebuilds for making compatible to any os type

PR #199, opened 2023-02-16
Weird solution

ZabFTFT:


WDemidenko (approved):
Very interesting solution, but might be a little too overcomplicated!

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    def __init__(self, path: str) -> None:
+        self.path = path
+
+    def __str__(self) -> str:
+        return f"Directory with path {self.path} is already exist!"
+
+
+def move_file(command: str) -> None:
+    input_command, path_file_in, path_file_out = command.split(" ")
```
YuriiKindrat, line 14: split method use spaces by default

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    if input_command != "mv":
+        return
+
+    if ("/" or "\\") not in path_file_out:
+        os.rename(path_file_in, path_file_out)
+        return
```
YuriiKindrat, line 21: try to avoid using blank return

app/main.py
```diff
@@ -1 +1,44 @@
...
+        print(e)
+    else:
+        os.makedirs(output_directory_path)
+
+    with (
+        open(absolute_file_in, "r") as file_in,
+        open(output_directory_file_path, "w") as file_out,
+    ):
+        read_content = file_in.read()
+        file_out.write(read_content)
```
YuriiKindrat, line 42: u can write this in one line

YuriiKindrat (changes requested):

app/main.py
ZabFTFT: Is it just bad practice? Do I have better option to terminate program if I want to? Like sys.exit ?

app/main.py
Serhii-Leonenko: fix this

app/main.py
```diff
@@ -1 +1,43 @@
...
+import os
+
+
+class DirectoryExistError(Exception):
+
+    def __init__(self, path: str) -> None:
+        self.path = path
+
+    def __str__(self) -> str:
+        return f"Directory with path {self.path} is already exist!"
```
Serhii-Leonenko, line 10: Use built-in exception https://docs.python.org/3/library/exceptions.html#FileExistsError

app/main.py
```diff
@@ -1 +1,43 @@
...
+
+
+def move_file(command: str) -> None:
+    input_command, path_file_in, path_file_out = command.split(" ")
+
+    if input_command != "mv":
+        exit()
```
Serhii-Leonenko, line 17: you can use `return` here, but also you can put cod inside this condition

app/main.py
```diff
@@ -1 +1,43 @@
...
+    else:
+        os.makedirs(output_directory_path)
+
+    with (
+        open(absolute_file_in, "r") as file_in,
+        open(output_directory_file_path, "w") as file_out,
+    ):
+        file_out.write(file_in.read())
+
+    os.remove(absolute_file_in)
```
Serhii-Leonenko, line 43: Your code is complicated, try to simplify it. 

HINT: you can use here `os.path.split() `

Serhii-Leonenko (changes requested):

app/main.py
ZabFTFT: Thank you for hint, crazy method. Next time I will learn the lib completely before start practice.

app/main.py
YuriiKindrat: why do u need this part of code, if 27-32 lines will make the same
u can just make a backward condition for code below, and use 27-32 lines for both conditions

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+    if ("/" or "\\") in path_file_out:
+        with (
+            open(path_file_in) as file_in,
+            open(output_directory_file_path, "w") as file_out,
+        ):
+            file_out.write(file_in.read())
+        os.remove(path_file_in)
+    else:
+        os.rename(path_file_in, path_file_out)
```
YuriiKindrat, line 31: it seems to me, this parts of code (24-29 lines and 31) line will make the same thing for command without " / " so why do u need os.rename ?

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #200, opened 2023-02-16
Solution

WDemidenko:


ZabFTFT (approved):
It's seems good but in checklist we have a recommendation to use os.path.join and it could give us utility to use code on cross-platform purpose.You did good job so I don't want to reject it cause of some suggestions.Well done!

YuriiKindrat (approved):

PR #201, opened 2023-02-17
Error solution

Rename925:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    cm, first_file, second_file = command.split()
```
ArsenPidhoretskyi, line 5: You code will not work if command will be one of:
"mv"
"mv file1.txt"
"mv file1.txt file2.txt file3.txt"

app/main.py
```diff
@@ -1 +1,20 @@
...
+    if cm == "mv" and first_file != second_file:
+        directory = second_file.split("/")
```

ArsenPidhoretskyi, line 7: It possible to be more than one directory

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,25 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    cm, first_file, second_file = command.split()
```
Vasyl-Smutok, line 5: you should add a check that `command.split()` is equal to `3`, only after that perform all actions


You code will not work if command will be one of:
"mv"
"mv file1.txt"
"mv file1.txt file2.txt file3.txt"

Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1 +1,26 @@
...
+    if len(command.split()) == 3:
+        cm, first_file, second_file = command.split()
+        if cm == "mv" and first_file != second_file:
+            if "/" not in second_file:
+                create_file(first_file, second_file)
+            else:
+                directory = second_file.split("/")[0:-1]
+                os.makedirs(os.path.join(*directory), exist_ok=True)
+                create_file(first_file, second_file)
+
```
Vasyl-Smutok, line 14: add `else` and raise some Exception

Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1 +1,28 @@
...
+        cm, first_file, second_file = command.split()
+        if cm == "mv" and first_file != second_file:
+            if "/" not in second_file:
+                create_file(first_file, second_file)
+            else:
+                directory = second_file.split("/")[0:-1]
+                os.makedirs(os.path.join(*directory), exist_ok=True)
+                create_file(first_file, second_file)
+    else:
+        raise Exception("wrong type of command")
```
katevasilenko, line 15: It’s a bad practice to raise just Exception. You need to specify it.

katevasilenko (changes requested):

vasylhnatiuk (approved):
Good job!

PR #203, opened 2023-02-18
'Solution'

alexlega:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    move_command, from_file, to_file = command.split()
```
SergiiMachulin, line 5: Why do you need a move_command?

app/main.py
```diff
@@ -1 +1,18 @@
...
+    with open(from_file) as fr_f:
+        content = fr_f.read()
```
SergiiMachulin, line 7: U could combine this context manager with the last one, and make you code conciser

app/main.py
```diff
@@ -1 +1,18 @@
...
+    dirs = to_file.split("/")
+    final_name = dirs.pop(-1)
+    path = ""
+    for directory in dirs:
+        path = os.path.join(path, directory)
+        if not os.path.exists(path):
+            os.mkdir(path)
```
SergiiMachulin, line 14: U can do this in two line with ```os.path.split``` and ```os.makedirs``` methods

SergiiMachulin (changes requested):
Code is OK, but as suggestion change in context managers as part and make it more descriptive according to the checklist)

app/main.py
alexlega: os.path.split could split path into "head" and "tail", but I anyway need to move through the path and check if does first directory in the path exist, and then check one-by-one directories. Of course, I still can use os.path.split, but I should split the path in the "head" part and make all the steps to check, which I explain before. But maybe I don't understand your proposal, sry

app/main.py
```diff
@@ -1 +1,19 @@
...
+
+
+def move_file(command: str) -> None:
+    move_command, from_file, to_file = command.split()
+    if move_command != "mv":
+        raise SyntaxError("No such command")
+    dirs = to_file.split("/")
+    final_name = dirs.pop(-1)
+    path = ""
+    for i in dirs:
```
YuriiKindrat, line 11: i is a bad name for var

app/main.py
```diff
@@ -1 +1,19 @@
...
+        path = os.path.join(path, i)
+        if not os.path.exists(path):
+            os.mkdir(path)
```
YuriiKindrat, line 14: u can use mkdirs method here, with exist_ok param

app/main.py
```diff
@@ -1 +1,19 @@
...
+    final_name_path = os.path.join(path, final_name)
+    with open(from_file) as fr_f, open(final_name_path, "w") as to_f:
+        content = fr_f.read()
```
YuriiKindrat, line 17: no need to create variable for this, u can write this directly into write

app/main.py
```diff
@@ -1 +1,19 @@
...
+        raise SyntaxError("No such command")
+    dirs = to_file.split("/")
+    final_name = dirs.pop(-1)
+    path = ""
+    for i in dirs:
+        path = os.path.join(path, i)
+        if not os.path.exists(path):
+            os.mkdir(path)
+    final_name_path = os.path.join(path, final_name)
+    with open(from_file) as fr_f, open(final_name_path, "w") as to_f:
```
YuriiKindrat, line 16: fr_f and to_f - use full names for vars

YuriiKindrat (changes requested):

PR #204, opened 2023-02-18
solution

usrofgh:


amber-marichi (approved):
very nice! 
have you considered to add few more checks where you throw the Exception,? like a case when the string in "file" variable ends with "/" , smthg like that?

app/main.py
```diff
@@ -1 +1,28 @@
-# write your code here
+from os import mkdir, remove, path, rename
+
+
+def move_file(command: str) -> None:
+    try:
+        mv, file, new_location = command.split()
```
YuriiKindrat, line 6: mv is a bad name for variable, use more descriptive names for naming

app/main.py
```diff
@@ -1 +1,28 @@
...
+        if mv != "mv":
+            raise ValueError
+    except ValueError:
+        raise ValueError(
+            "Incorrect command. Example: mv file1.txt file2.txt "
+            "or mv file1.txt new_folder/file1.txt"
+        )
```
YuriiKindrat, line 13: why do u need to use try except to raise the same exception ?

app/main.py
```diff
@@ -1 +1,28 @@
...
+    if "/" not in new_location and file != new_location:
+        rename(file, new_location)
+
+    elif "/" in new_location:
+        directories, new_file = new_location.rsplit("/", 1)
+        full_path = ""
+        for directory in directories.split("/"):
+            full_path = path.join(full_path, directory)
+            if not path.exists(full_path):
+                mkdir(full_path)
```
YuriiKindrat, line 24: try to use mkdirs method to siplify ur code

app/main.py
```diff
@@ -1 +1,28 @@
...
+        if mv != "mv":
+            raise ValueError
+    except ValueError:
+        raise ValueError(
+            "Incorrect command. Example: mv file1.txt file2.txt "
+            "or mv file1.txt new_folder/file1.txt"
+        )
+
+    if "/" not in new_location and file != new_location:
+        rename(file, new_location)
```
YuriiKindrat, line 16: why do u need this part of code, if 26-28 lines will do the same?

YuriiKindrat (changes requested):

app/main.py
usrofgh: If i delete this part then the program will occur an error, cause if we can only file, without directories, then during splitting in 21 line we'll get an error. [Demonstration](https://prnt.sc/OUQYXXlTtvUG)

file.txt
```diff
@@ -0,0 +1,3 @@
+This is some
+ content for
+ the file.
```
YuriiKindrat, line 3: remove unnecessary file from ur PR

app/main.py
```diff
@@ -1 +1,20 @@
...
+
+    if "/" not in new_location:
+        rename(file, new_location)
+    else:
+        directories, new_file = new_location.rsplit("/", 1)
+        makedirs(directories)
+
+        with open(file) as read_file, open(new_location, "w") as write_file:
+            write_file.write(read_file.read())
+        remove(file)
```
YuriiKindrat, line 20: 13 and 18-20 lines will make the same thing, so make ur condition in such way, so there is won't need to use rename func

YuriiKindrat (changes requested):

wQuelS (changes requested):

YuriiKindrat (approved):

PR #205, opened 2023-02-18
Solution

yura-hudzovskyi:


inrock1 (approved):

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command, file_before, direction = command.split()
```
YuriiKindrat, line 5: image
u should check if command is correct

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,18 @@
...
+    direction = direction.split("/")
```
Serhii-Leonenko, line 6: Is your app cross-platform?) For example, will it works with `\` on windows ?)

HINT: `use os.path.split()`

app/main.py
```diff
@@ -1 +1,18 @@
...
+    new_file = direction.pop(-1)
+    path = ""
+    if command != "mv":
+        raise ValueError("Wrong command")
+    for directory in direction:
+        path += directory + "/"
+        if not os.path.exists(path):
+            os.mkdir(path)
```
Serhii-Leonenko, line 14: Can you try to use `os.makedirs()` without loop?)

app/main.py
```diff
@@ -1 +1,18 @@
...
+
+    with open(file_before) as old_file, open(path + new_file, "w") as file:
```
Serhii-Leonenko, line 16: what about `mode` in the firs open?

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #206, opened 2023-02-18
Solution

Sashkoooo:


app/main.py
```diff
@@ -1 +1,31 @@
...
+        source_folder = ""
+
+        if command != "mv":
+            raise ValueError("Incorrect command")
+
+        for folder in folders:
+            destination_path = os.path.join(source_folder, folder)
+            source_folder = destination_path
+            if not os.path.exists(destination_path):
+                os.mkdir(destination_path)
```
ArsenPidhoretskyi, line 18: You can use:
os.path.join(*folders)
and use os.makedirs with exist_ok=True

app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    try:
+        command, file_to_move, path = command.split()
+        folders = path.split("/")
+        new_file_name = folders.pop(-1)
+        source_folder = ""
```
ArsenPidhoretskyi, line 9: source_folder is not clear name because you use it as destination

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #207, opened 2023-02-19
Solution

SergiiMachulin:


app/main.py
```diff
@@ -1 +1,25 @@
...
+
+def move_file(command: str) -> None:
+    if len(command.split(" ")) != 3 or command.split(" ")[0] != "mv":
+        return
+    command_parts = command.split(" ")
+
+    if os.path.exists(command_parts[1]):
+        file_in_name = command_parts[1]
+    else:
+        raise FileNotFoundError("No such file you try to move")
```
YuriiKindrat, line 13: no need to use else here, u can make a condition just to raise Exception

app/main.py
```diff
@@ -1 +1,25 @@
...
+
+    if command_parts[2].count("/") == 0:
```
YuriiKindrat, line 15: no need to write == 0 here

YuriiKindrat (changes requested):

wQuelS (approved):
GJ

PR #208, opened 2023-02-19
solution

amber-marichi:


oksana-feshchenko (approved):
Greate job!

app/main.py
```diff
@@ -1 +1,50 @@
...
+        operation, old_file_loc, new_file_loc = command.split()
+        if not os.path.exists(old_file_loc) or os.path.isdir(old_file_loc):
+            raise ValueError
+    except ValueError:
+        print(
+            "Please check the input, "
+            "it should be like 'mv file some_dir/new_file'. "
+            "The file to move cannot be directory!"
+        )
+        return
```
YuriiKindrat, line 26: smplify ur code try to use backward condition for this case

YuriiKindrat (changes requested):
image
fix this please

app/main.py
```diff
@@ -1 +1,38 @@
...
+        if "/" in new_file_loc:
+            new_path, *_ = new_file_loc.rsplit("/", 1)
+            if new_file_loc.endswith("/"):
+                if "/" in old_file_loc:
+                    _, old_file_name = old_file_loc.rsplit("/", 1)
+                    new_file_loc = new_path + "/" + old_file_name
+                else:
+                    new_file_loc = new_path + "/" + old_file_loc
+
+            _check_path(new_path)
```
Serhii-Leonenko, line 34: Will it works with the `\` instead of `/` ?) Can you try to simplify and make universally your code? You can use os.path.split(), os.makedirs() and other built-in functions to do this.

app/main.py
```diff
@@ -1 +1,38 @@
...
+
+        with open(old_file_loc) as reader, open(new_file_loc, "w") as writer:
```
Serhii-Leonenko, line 36: What about  "mode" in reader open ?

Serhii-Leonenko (changes requested):

app/main.py
amber-marichi: great tip! tyvm

app/main.py
```diff
@@ -1 +1,38 @@
...
+        operation, old_file_loc, new_file_loc = command.split()
+        if not os.path.isfile(old_file_loc):
+            raise ValueError
+    except ValueError:
+        print(
+            "Please check the input, "
+            "it should be like 'mv file some_dir/new_file'. "
+            "The file to move cannot be directory!"
+        )
+        return
```
Serhii-Leonenko, line 23: For what do you raise and then handle the error?)

app/main.py
```diff
@@ -1 +1,38 @@
...
+    if operation == "mv" and old_file_loc != new_file_loc:
+        if os.path.isdir(new_file_loc):
+            new_path, new_file_name = new_file_loc, None
+        else:
+            new_path, new_file_name = os.path.split(new_file_loc)
```
Serhii-Leonenko, line 28: are you sure that you need this condition?)

app/main.py
```diff
@@ -1 +1,38 @@
...
+import os
+
+
+def _check_path(new_path: str) -> None:
+    cur_path = ""
+    for directory in new_path.split("/"):
+        cur_path = os.path.join(cur_path, directory)
+        if not os.path.exists(cur_path):
+            os.mkdir(cur_path)
+
```
Serhii-Leonenko, line 10: you can remake this check, try to use os.makedirs() and remove for loop

Serhii-Leonenko (changes requested):
Try to remove redundant conditions, loops, etc.

app/main.py
amber-marichi: this was check in case when split() returns more or less variables than we need. Also i wanted to raise the same error in case when name of directory was passed as a file to move. nvm i removed it

app/main.py
amber-marichi: tried to make this as a check for existing directories in cases of multiple depths. replaced it with os.makedirs()

app/main.py
```diff
@@ -1 +1,15 @@
...
+def move_file(command: str) -> None:
+    operation, old_file_loc, new_file_loc = command.split()
+
+    if operation == "mv" and old_file_loc != new_file_loc:
+        new_path, _ = os.path.split(new_file_loc)
+        if new_path:
+            os.makedirs(new_path, exist_ok=True)
+
+        with open(old_file_loc, "r") as reader:
+            with open(new_file_loc, "w") as writer:
```
YuriiKindrat, line 13: no need to use "with" for 2 times
image

YuriiKindrat (approved):

PR #209, opened 2023-02-19
'Solution'

inrock1:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if len(command.split()) != 3 or command[:3] != "mv ":
+        return 0
```
viktoria-rybenchuk, line 6: I think you can return without 0

viktoria-rybenchuk (approved):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command, file_in, file_out_with_rout = command.split()
+    if command == "mv":
+        rout_list = file_out_with_rout.split("/")
+        file_out_with_rout_cp = os.path.join(*rout_list)
```
Serhii-Leonenko, line 8: Do you really need this?
Will it work without these two lines of code?

Serhii-Leonenko (changes requested):

app/main.py
inrock1: image
this is from checklist for this task

Serhii-Leonenko: You need this if you don't use os.renames(). For example, you separately create directories and rewrite the file. But in this case your code will work with string like this
`move_file("mv file.txt first_dir/second_dir/third_dir/file2.txt")`
also with string like this
move_file("mv file.txt first_dir\\second_dir\\third_dir\\file2.txt")

YuriiKindrat (approved):

PR #211, opened 2023-02-20
Develop

BohdanLazaryshyn:


yura-hudzovskyi (approved):

app/main.py
```diff
@@ -1 +1,17 @@
...
+
+def move_file(command: str) -> None:
+    command_sep = command.split()
+    if command_sep[0] == "mv":
+        old_name = command_sep[1]
+        new_name = command_sep[-1].split("/")[-1]
+        directory = command_sep[-1].split("/")
+        way = ""
+        if len(directory) > 1:
+            way = "/".join(directory[0: -1: 1]) + "/"
```
wQuelS, line 12: what if your program runs on another system where the path is delimited by `\` ?
use `os.path.join()` from module that u've imported in the first line.

Also think about how to improve your logic, 'cause this slicing is not the optimal solution

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_sep = command.split()
+    if command_sep[0] == "mv":
+        old_name = command_sep[1]
+        new_name = command_sep[-1].split("/")[-1]
+        directory = command_sep[-1].split("/")
```
wQuelS, line 9: reimplement this logic, since such hardcode is not good. Remember that you can `split()` like this:
`var, var1, var2 = some_string.split()`

app/main.py
```diff
@@ -1 +1,17 @@
...
+        way = ""
+        if len(directory) > 1:
+            way = "/".join(directory[0: -1: 1]) + "/"
+            if not os.path.exists(way):
+                os.makedirs(way)
+        with open(old_name, "r") as old, open(way + new_name, "w") as new:
```
wQuelS, line 15: concatenation in `open` is not good either

wQuelS (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    _, old_name, directory = command.split()
```
wQuelS, line 5: You are using `_` right below this line. So better give it a name. Cause we usually name variables as `_` when we are sure that we never ever gonna use it

wQuelS (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
...
+
+    if flag == "mv":
+        path, new_file = os.path.split(directory)
+
+        if not os.path.exists(path) and path != "":
+            os.makedirs(name=path)
+
+        with open(old_name, "r") as old:
+            with open(directory, "w") as new:
+                new.write(old.read())
```
YuriiKindrat, line 15: image

app/main.py
```diff
@@ -1 +1,17 @@
...
+
+
+def move_file(command: str) -> None:
+    flag, old_name, directory = command.split()
+
+    if flag == "mv":
+        path, new_file = os.path.split(directory)
+
+        if not os.path.exists(path) and path != "":
+            os.makedirs(name=path)
```
YuriiKindrat, line 11: u can use "exist_ok" param for makedirs func

YuriiKindrat (changes requested):

PR #218, opened 2023-02-22
v1

MaxiTDI:


app/main.py
```diff
@@ -1 +1,35 @@
...
+
+
+def move_file(command: str) -> None:
+    validate(command)
+    cmd, old_file_name, directory_file = command.split()
+    if "/" not in directory_file:
+        os.rename(old_file_name, directory_file)
+    else:
+        path_list = directory_file.split("/")
+        path = "/".join(path_list[:-1])
```
ArsenPidhoretskyi, line 24: Use os.path split, also use os.path.join

app/main.py
```diff
@@ -1 +1,35 @@
...
+        file_name = path_list[-1]
+        if not os.path.exists(path):
+            os.makedirs(path)
```
ArsenPidhoretskyi, line 27:
```suggestion
        os.makedirs(path, exists_ok=True)
```

app/main.py
```diff
@@ -1 +1,35 @@
...
+        try:
+            with open(old_file_name) as old_file, \
+                    open(os.path.join(path, file_name), "w") as new_file:
```
ArsenPidhoretskyi, line 30:
```suggestion
            with (
                open(old_file_name) as old_file,
                open(os.path.join(path, file_name), "w") as new_file
            ):
```

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Good job!

PR #221, opened 2023-02-22
Solution

Esxoyne:


app/main.py
```diff
@@ -1 +1,28 @@
...
+        return
+    if not os.path.exists(source):
+        raise FileNotFoundError("Source file does not exist")
+
+    if "/" not in destination:
+        os.rename(source, destination)
+        return
+
+    directories = destination.split("/")[:-1]
+    path = ""
```
ArsenPidhoretskyi, line 20: you can use os.path.split
Usage of .split("/") is not correct because on different os it will works differently

app/main.py
```diff
@@ -1 +1,28 @@
...
+    for directory in directories:
+        path = os.path.join(path, directory)
+        if not os.path.exists(path):
+            os.mkdir(path)
```
ArsenPidhoretskyi, line 24: use:
 os.path.join(*list_of_dirs)
os.makedirs with exist_ok=True

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,23 @@
...
+    if command != "mv" or source == destination:
+        return
+    if not os.path.exists(source):
+        raise FileNotFoundError("Source file does not exist")
+
+    if "/" not in destination:
+        os.rename(source, destination)
+        return
+
+    directories = os.path.split(destination)[0]
```
ArsenPidhoretskyi, line 19: destination will return two:
directories and file_name

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,24 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    try:
+        command, source, destination = command.split()
+    except ValueError:
+        print("Three arguments expected")
```
ArsenPidhoretskyi, line 8: use raise instead of print

app/main.py
```diff
@@ -1 +1,24 @@
...
+    if command != "mv" or source == destination:
+        return
+    if not os.path.exists(source):
+        raise FileNotFoundError("Source file does not exist")
+
+    directories, file_name = os.path.split(destination)
+    if not directories:
+        os.rename(source, destination)
+        return
+    os.makedirs(directories, exist_ok=True)
```
ArsenPidhoretskyi, line 19: directories is list of str
you need to join it before

ArsenPidhoretskyi (changes requested):

Grayder0152 (approved):

PR #222, opened 2023-02-23
Solution

Zenionoi:
In the test I accidentally added an empty line

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,26 @@
-# write your code here
+import os
+import shutil
+
+
+def move_file(command: str) -> None:
+
+    command_split = command.split()
+    if (command_split[0] != "mv"
+            or command_split[1][-4:] != command_split[2][-4:]
+            or command_split[1][-4:] != ".txt"):
+        raise ValueError("Error in entered data!")
+
+    *directory, new_file = command_split[-1].split("/")
+
+    if not directory:
+        os.rename(command_split[1], new_file)
+    else:
+        path = ""
+        for folder in directory:
+            path += folder + "/"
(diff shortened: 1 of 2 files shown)


tests/test_main.py
```diff
@@ -27,6 +27,7 @@ def test_file_renamed(create_file: callable) -> None:
 
 
 def test_should_work_when_directory_exists(create_file: callable) -> None:
+
```
wQuelS, line 30: do not change tests! fix it

app/main.py
```diff
@@ -1 +1,26 @@
...
+import os
+import shutil
+
+
+def move_file(command: str) -> None:
+
+    command_split = command.split()
+    if (command_split[0] != "mv"
+            or command_split[1][-4:] != command_split[2][-4:]
+            or command_split[1][-4:] != ".txt"):
```
wQuelS, line 10: bad implementation. Can you try to make another solution without such hardcoding?

app/main.py
```diff
@@ -1 +1,26 @@
...
+        raise ValueError("Error in entered data!")
+
+    *directory, new_file = command_split[-1].split("/")
+
+    if not directory:
+        os.rename(command_split[1], new_file)
+    else:
+        path = ""
+        for folder in directory:
+            path += folder + "/"
```
wQuelS, line 20: what if program runs on another OS which uses `\` as path delimiter?
use `os.path.join()` from module that u've already imported

wQuelS (changes requested):

YuriiKindrat (approved):

PR #223, opened 2023-02-24
solved task 1

SDobzhynetskyi:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+import os
+import shutil
+
+
+def move_file(command: str) -> None:
+    operation, file_in, file_to_move = command.split()
```
ArsenPidhoretskyi, line 6: Need to validate that command have exactly the 3 parameters

app/main.py
```diff
@@ -1 +1,20 @@
...
+    if operation == "mv" and "/" not in file_to_move:
```
checking for `"/"` is not correct way

app/main.py
```diff
@@ -1 +1,20 @@
...
+        os.rename(file_in, file_to_move)
+    if operation == "mv" and "/" in file_to_move:
+        path_generator = file_to_move.split("/")
```
ArsenPidhoretskyi, line 10: the same with os.path.split

app/main.py
```diff
@@ -1 +1,20 @@
...
+        folder_index = 0
+        path_for_creation = ""
+        while folder_index < len(path_generator) - 1:
+            path_for_creation += f"{path_generator[folder_index]}/"
+            try:
+                os.mkdir(path_for_creation)
+            except FileExistsError:
+                pass
+            folder_index += 1
```
ArsenPidhoretskyi, line 19: You can use os.path.join(*list_of_directories)
os.makedirs with flag exist_ok=True

app/main.py
```diff
@@ -1 +1,20 @@
...
+        shutil.move(file_in, f"{path_for_creation}{path_generator[-1]}")
```
ArsenPidhoretskyi, line 20: Not need to use libraries like shututil
You must implement custom realization

ArsenPidhoretskyi (changes requested):

app/main.py
ArsenPidhoretskyi: Not fixed need to validate it before operation, file_in, file_to_move = command.split()

ArsenPidhoretskyi: Pin a screenshot, if your code will work with following(must raise error with description):
cp file1.txt
cp file1.txt file2.txt file3.txt

app/main.py
```diff
@@ -1 +1,24 @@
...
+            or operation != "mv"
+            or not os.path.isfile(file_in)
+            or file_to_move.endswith("/")
+    ):
+        raise ValueError("The input command has wrong structure")
+    if not os.path.exists(file_to_move):
+        try:
+            os.rename(file_in, file_to_move)
+        except FileNotFoundError:
+            path, new_file = os.path.split(file_to_move)
```
ArsenPidhoretskyi, line 17: path, new_file is not clear names
path -> directories
new_file -> file_name

ArsenPidhoretskyi (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #228, opened 2023-02-26
Add solution

ZaikaDaria:


app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    cmd, filename, destination_path_with_file = command.split()
```
Grayder0152, line 5: Beforeif pls check that `len(command) == 3`

app/main.py
```diff
@@ -1 +1,13 @@
...
+    if cmd == "mv" and filename != destination_path_with_file:
+        os.makedirs(os.path.dirname(destination_path_with_file))
```
Grayder0152, line 7: What if dirs already exists?

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if len(command) != 3:
```
ArsenPidhoretskyi, line 5: The wrong check!
command is a string
It works incorrectly. Did you test your code manually? 
For example "mv file1.txt file2.txt"

app/main.py
```diff
@@ -1 +1,17 @@
...
+            os.makedirs(
+                os.path.dirname(destination_path_with_file),
+                exist_ok=True
+            )
+            with (
+                open(f"{filename}", "r") as file_in,
+                open(f"{destination_path_with_file}", "w") as file_out
+            ):
+                file_out.write(file_in.read())
+                os.remove(filename)
```
ArsenPidhoretskyi, line 17: move this after context manager

ArsenPidhoretskyi (changes requested):

Grayder0152 (approved):

PR #230, opened 2023-02-27
Solution

yhwh6:


app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    source_path, dest_path = command.split()[1:]
+    if "/" not in dest_path:
```
ArsenPidhoretskyi, line 6: Before use os.path.split and then check due to directory path

app/main.py
```diff
@@ -1 +1,14 @@
...
+        os.rename(source_path, dest_path)
+    elif "/" in dest_path:
+        dir_names = []
+        for name in dest_path.split("/")[:-1]:
+            dir_names.append(name)
+            if not os.path.exists("/".join(dir_names)):
+                os.mkdir("/".join(dir_names))
```
ArsenPidhoretskyi, line 13: Use os.makedirs(directories_path, exist_ok=True)

app/main.py
```diff
@@ -1 +1,14 @@
...
+        os.rename(source_path, dest_path)
```
ArsenPidhoretskyi, line 14: Move it after if due to DRY

ArsenPidhoretskyi (changes requested):
image
https://github.com/mate-academy/py-move-file/blob/master/checklist.md
You have to do the tasks as in real work, you are not just given a checklist)

vasylhnatiuk (approved):
Good job!

PR #232, opened 2023-03-01
solution

Bloodviel:


app/main.py
```diff
@@ -1 +1,23 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    cmd = command.split()
```
Grayder0152, line 5: Before it, pls check that `len(cmd)==3`

Grayder0152 (changes requested):

app/main.py
ArsenPidhoretskyi: "mv" in cmd not correct check
because:
commad == "cp filemv.txt file2.txt"

app/main.py
```diff
@@ -1 +1,23 @@
...
+    if "mv" in cmd and len(cmd) == 3:
+        path = cmd[cmd.index("mv") + 1:]
```
ArsenPidhoretskyi, line 7: You can use split and unpacking
input_command, source_file_path, destination_file_path = cmd

app/main.py
```diff
@@ -1 +1,23 @@
...
+        old_file = path[0]
+        if "/" in path[1]:
```
ArsenPidhoretskyi, line 9: Use os.path.split and then check due to dirs

ArsenPidhoretskyi (changes requested):

Grayder0152 (approved):

PR #233, opened 2023-03-01
solution

Yulia-Mala:


app/main.py
```diff
@@ -1 +1,16 @@
...
+from os import makedirs, path
+from shutil import move
+
+
+def move_file(command: str) -> None:
+    args = command.split()
+    source = args[1]
+    destination = args[2].split("/")
+    new_filename = destination[-1]
+    dirs = []
```
YuriiKindrat, line 10: use split to unpack values directly to the variables

app/main.py
```diff
@@ -1 +1,16 @@
...
+    if len(destination) > 1:
+        dirs = destination[:-1]
+        if not path.exists("/".join(dirs)):
+            makedirs("/".join(dirs))
+    dirs.append(new_filename)
+    move(source, "/".join(dirs))
```
YuriiKindrat, line 16: use os.path.join for instead of "/".join(dirs), also u can use exist_ok param for makefis func

YuriiKindrat (changes requested):

PR #235, opened 2023-03-03
'Solution'

natalie-goriela:


app/main.py
```diff
@@ -1 +1,28 @@
-# write your code here
+import os
+import shutil
+
+
+def move_file(command: str) -> None:
+    command = command.split(" ")
+    if "mv" not in command or len(command) > 3:
```
ArsenPidhoretskyi, line 7: len(command) != 3

app/main.py
```diff
@@ -1 +1,28 @@
...
+        return
+
+    if "/" not in command[2]:
```
ArsenPidhoretskyi, line 10: use os.path.split
Checks for "/" is so bad

app/main.py
```diff
@@ -1 +1,28 @@
...
+        os.rename(command[1], command[2])
+        return
+
+    if os.path.exists(command[2]):
+        return
+
+    file_name = command[1]
+    new_file_name = command[2].split("/").pop()
```
ArsenPidhoretskyi, line 18: os.path.split

app/main.py
```diff
@@ -1 +1,28 @@
...
+    path = command[2].replace(new_file_name, "")
```
ArsenPidhoretskyi, line 19: this will be redundant after usage of os.path.split

app/main.py
```diff
@@ -1 +1,28 @@
...
+
+    os.rename(file_name, new_file_name)
+
+    try:
+        os.makedirs(path)
+    except FileExistsError:
```
ArsenPidhoretskyi, line 25: Use os.makedirs(exist_ok=True)

app/main.py
```diff
@@ -1 +1,28 @@
...
+        shutil.move(new_file_name, path)
+    else:
+        shutil.move(new_file_name, path)
```
ArsenPidhoretskyi, line 28: Use custom realisation not need to use shutil

ArsenPidhoretskyi: Also it is duplicate of code

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,30 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command = command.split(" ")
```
olena-hul, line 5:
```suggestion
    command = command.split()
```

app/main.py
```diff
@@ -1 +1,30 @@
...
+    if "mv" not in command or len(command) != 3:
+        return
+
+    if "/" not in command[2]:
+        os.rename(command[1], command[2])
```
olena-hul, line 10: Not fixed
Use unpacking (for example `a, b = 3, 4`) instead of indexes to access command parts

olena-hul (changes requested):

Bohdan-Salamakha (approved):

PR #239, opened 2023-03-10
/

DmytroProkofiev:


app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if command.split()[0] == "mv" and command.split() == 3:
```
ArsenPidhoretskyi, line 5: swap and add len because it return list
```suggestion
    if len(command.split()) == 3 command.split()[0] == "mv":
```

app/main.py
```diff
@@ -1 +1,19 @@
...
+        command, source_file_name, new_file_info = command.split()
+        if "/" not in new_file_info:
+            new_file_name = new_file_info
+        else:
+            directories, file_name = os.path.split(new_file_info)
+            if directories:
+                os.makedirs(directories, exist_ok=True)
+            new_file_name = os.path.join(directories, file_name)
```
ArsenPidhoretskyi, line 13: You can remove if-else here
```suggestion
        directories, file_name = os.path.split(new_file_info)
        if directories:
            os.makedirs(directories, exist_ok=True)
        new_file_name = new_file_info # attention here you can remove this line
```

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,15 @@
...
+        command, source_file_name, new_file_info = command.split()
+        directories, file_name = os.path.split(new_file_info)
+        if directories:
+            os.makedirs(directories, exist_ok=True)
+        new_file_name = os.path.join(directories, file_name)
+        with open(f"{source_file_name}", "r") as old_file:
+            temp_text = old_file.read()
+        os.remove(source_file_name)
+        with open(f"{new_file_name}", "w") as new_file:
+            new_file.write(temp_text)
```
Bohdan-Salamakha, line 15: open two streams at the same time

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #244, opened 2023-03-15
'Solution'

StinGER-ua:


app/main.py
```diff
@@ -1 +1,24 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command = command.split()
+    if command[0] == "mv" and command[2][-1] != "/":
```
ArsenPidhoretskyi, line 6: validate that command have 3 params
also it will does not work on windows: command[2][-1] != "/"

app/main.py
```diff
@@ -1 +1,24 @@
...
+
+
+def move_file(command: str) -> None:
+    command = command.split()
+    if command[0] == "mv" and command[2][-1] != "/":
+
+        with open(command[1]) as source:
+            data = source.read()
+
+        path = command[2].split("/")
```
ArsenPidhoretskyi, line 11: use os.path.split

app/main.py
```diff
@@ -1 +1,24 @@
...
+        directory = ""
+
+        for folder in path[:-1]:
+            new_path = os.path.join(directory, folder)
+            if not os.path.isdir(new_path):
+                os.mkdir(new_path)
+            directory = new_path
```
ArsenPidhoretskyi, line 18: No need cycle use os.path.join(*path[:-1]) and than os.makedirs

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    cmd, source_file, final_file = command.split()
+
+    if (
+            len(command.split()) == 3
```
ArsenPidhoretskyi, line 8: You need to validate it before 5 line
Check your code with command == "mv"

app/main.py
```diff
@@ -1 +1,31 @@
...
+import os
+
+
+def move_file(command: str) -> None:
+    cmd, source_file, final_file = command.split()
+
+    if (
+            len(command.split()) == 3
+            and cmd == "mv"
+            and final_file[-1] != "/"
```
ArsenPidhoretskyi, line 10: You do not need to check by "/", because on others OS will be another slash

app/main.py
```diff
@@ -1 +1,31 @@
...
+    ):
+
+        with open(source_file) as source:
+            data = source.read()
+
+        path, final_file_name = os.path.split(final_file)
```
ArsenPidhoretskyi, line 16: rename path to directories

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+        if len(path) > 1:
+            directory = os.path.join(path)
```
ArsenPidhoretskyi, line 19: path is strign why you need join here?

app/main.py
```diff
@@ -1 +1,31 @@
...
+        else:
+            directory = "."
+
+        if not os.path.exists(directory):
+            os.makedirs(directory)
```
ArsenPidhoretskyi, line 24: use exist_ok flag, no need to ckeck

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+        new_file = os.path.join(directory, final_file_name)
```
ArsenPidhoretskyi, line 26: Redundant code

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+        with open(new_file, "w") as copy:
```
ArsenPidhoretskyi, line 28:
```suggestion
        with open(final_file, "w") as copy:

```

ArsenPidhoretskyi: copy is not correct name. rename it to destination_file

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,28 @@
...
+
+        if len(final_path) > 1:
+            directory = final_path
+        else:
+            directory = "."
+
+        if not os.path.exists(directory):
+            os.makedirs(directory, exist_ok=True)
+
+        with open(os.path.join(directory, fin_file_name), "w") as target_file:
```
ArsenPidhoretskyi, line 25: And then here just do 
```suggestion
        with open(final_file, "w") as target_file:

```

ArsenPidhoretskyi (changes requested):

app/main.py
StinGER-ua: I got confused

app/main.py
```diff
@@ -1 +1,23 @@
...
+        with open(source_file) as source:
+            data = source.read()
+
+        final_path, fin_file_name = os.path.split(final_file)
+
+        if final_path:
+            os.makedirs(final_path, exist_ok=True)
+
+        with open(final_file, "w") as target_file:
+            target_file.write(data)
```
Grayder0152, line 21:
```suggestion
        final_path, fin_file_name = os.path.split(final_file)

        if final_path:
            os.makedirs(final_path, exist_ok=True)

        with (
            open(final_file, "w") as target_file,
            open(source_file) as source
        ):
            data = source.read()
            target_file.write(data)
```

Grayder0152 (changes requested):

Grayder0152 (approved):

PR #245, opened 2023-03-18
Solution

oleg-potichnyi:


app/main.py
```diff
@@ -1 +1,23 @@
...
+        dst_file = destination_path
+    os.replace(source_path, dst_file)
+
+
+with open("file.txt", "w") as source:
+    source.write("Some\nText\n")
+move_file("mv file.txt first_dir/second_dir/third_dir/file2.txt")
+with open("first_dir/second_dir/third_dir/file2.txt", "r") as new_files:
+    print(new_files.read())
+os.remove("first_dir/second_dir/third_dir/file2.txt")
```
Bohdan-Salamakha, line 23: remove it

Bohdan-Salamakha (changes requested):
u need to create a directory if it doesn't exists
image

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    parts = command.split()
+    source_path = parts[1]
```
ArsenPidhoretskyi, line 6: You need to validate that command have exactly 3 arguments

app/main.py
```diff
@@ -1 +1,17 @@
...
+    destination_path = parts[2]
+    if not os.path.isfile(source_path):
+        raise FileNotFoundError(f"Source file {source_path} not found")
+    if destination_path.endswith("/"):
```
ArsenPidhoretskyi, line 10: Use os.path.split

ArsenPidhoretskyi: Also code is not works when dir in path but not ended by /

ArsenPidhoretskyi (commented):
Also dst_file is redundant variable delete it

ArsenPidhoretskyi (changes requested):
Changes requested

app/main.py
oleg-potichnyi: "Also code is not works when dir in path but not ended by /"  Please explain this in more detail

app/main.py
olena-hul: This is not fixed

app/main.py
olena-hul: For example when I use the following command
`mv file.txt folder`

I expect the file to be moved to `folder` directory
But the code won't work because you assume that directory ends exactly with `/` symbol but it is not always true

olena-hul (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
...
+import os
+
+
+def move_file(command: str) -> None:
+    parts = command.split()
+    source_path = parts[1]
+    if len(parts) != 3:
+        raise ValueError("Invalid format")
+    destination_path = parts[2]
+    if not os.path.split(source_path) and destination_path.endswith("/"):
```
olena-hul, line 10: The check is wrong

olena-hul: Test this code with destination path that doesn't end with `/` - it will fail
We should allow such destination paths either.

And I don't understand why did you add this `not os.path.split(source_path)`

olena-hul (changes requested):

Bohdan-Salamakha (changes requested):
first test fails
https://github.com/mate-academy/py-move-file#:~:text=mv%20file.txt%20file2.txt%20simply%20renames%20the%20file.
image

oleg-potichnyi:
I can't understand what the error is. I ask for a more detailed explanation.

Bohdan-Salamakha:
> I can't understand what the error is. I ask for a more detailed explanation.

Script doesn't rename the file as needed in example

Bohdan-Salamakha (changes requested):
v3

oleg-potichnyi:
Why does this happen, I used the os.rename() method?

Bohdan-Salamakha:
> Why does this happen, I used the os.rename() method?

Check it all in debug

Bohdan-Salamakha (changes requested):
v4

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if len(command.split()) != 3:
+        raise ValueError("Invalid format")
+    md, source_path, destination_path = command.split()
```
Bohdan-Salamakha, line 7: u call command.split() twice
don't do it
create separated variable for command.split() and use it when u need it

Bohdan-Salamakha (changes requested):
v2

app/main.py
```diff
@@ -1 +1,17 @@
...
+    parts = command.split()
+    if len(parts) != 3:
+        raise ValueError("Invalid format")
+    md, source_path, destination_path = parts
+    if md not in command or source_path == destination_path:
+        raise FileNotFoundError(f"Source file {source_path} not found")
+    if not os.path.isfile(source_path):
+        raise FileNotFoundError(f"Source file {source_path} not found")
+    directories_path, file_name = os.path.split(source_path)
+    if not len(directories_path) != 0:
```
ArsenPidhoretskyi, line 14:
```suggestion
    if len(directories_path):
```

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
...
+        os.makedirs(directories_path, exist_ok=True)
+        os.path.join(directories_path, os.path.basename(source_path))
```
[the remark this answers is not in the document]
oleg-potichnyi: Which line to delete or the entire block of code?

olena-hul (changes requested):

app/main.py
```diff
@@ -1 +1,16 @@
...
+
+def move_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3:
+        raise ValueError("Invalid format")
+    md, source_path, destination_path = parts
+    if md not in command or source_path == destination_path:
+        raise FileNotFoundError(f"Source file {source_path} not found")
+    if not os.path.isfile(source_path):
+        raise FileNotFoundError(f"Source file {source_path} not found")
```
Bohdan-Salamakha, line 12:
```suggestion
    if md not in command or source_path == destination_path or not os.path.isfile(source_path):
        raise FileNotFoundError(f"Source file {source_path} not found")
```

app/main.py
```diff
@@ -1 +1,14 @@
...
+import os
+
+
+def move_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3:
+        raise ValueError("Invalid format")
+    md, source_path, destination_path = parts
+    if md not in command or source_path == destination_path or not os.path.isfile(source_path):
+        raise FileNotFoundError(f"Source file {source_path} not found")
```
Bohdan-Salamakha, line 10:
```suggestion
    if (
        md not in command or
        source_path == destination_path or
        not os.path.isfile(source_path)
):
    raise FileNotFoundError(f"Source file {source_path} not found")

```

Bohdan-Salamakha (changes requested):
fix flake8 a  bit and it will be great

Bohdan-Salamakha (approved):

PR #246, opened 2023-03-20
Solution

Anyta17:


app/main.py
```diff
@@ -1 +1,16 @@
...
+import os
+
+
+def move_file(command: str) -> None:
+    if command.split()[0] == "mv" and len(command.split()) == 3:
+        command, source_file_name, new_file_info = command.split()
+        directories, file_name = os.path.split(new_file_info)
+        if directories:
+            os.makedirs(directories, exist_ok=True)
+        new_file_name = os.path.join(directories, file_name)
```
ArsenPidhoretskyi, line 10: This is redundant you already have new_file_info

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if len(command.split()) == 3 and command.split()[0] == "mv":
```
olena-hul, line 5: `command.split()` can be moved to a variable

app/main.py
```diff
@@ -1 +1,15 @@
...
+        command, source_file_name, new_file_info = command.split()
```
olena-hul, line 6: Why did you name the variables so differently? `source_file_name` and `new_file_info`?
Use `source_file_name` and `target_file_name`(for example) instead to keep consistency

olena-hul (changes requested):

PR #250, opened 2023-03-27
'Solution'

iishchenko:


vasylhnatiuk (changes requested):

iishchenko:
Changes
1)i is a bad name
2)also you can use join instead of for loop - it will not work in this case: if the root folder in the destination path does not exist then the inner folder cannot be created

iishchenko:
Changes
1)i is a bad name
2)also you can use join instead of for loop - it will not work in this case: if the root folder in the destination path does not exist then the inner folder cannot be created

app/main.py
```diff
@@ -1 +1,23 @@
...
+            if len(destination_dirs) > 1:
+                temporary_directory = ""
+                for folder_name in destination_dirs[:-1]:
+                    temporary_directory = os.path.join(
+                        temporary_directory, folder_name
+                    )
+                    if not os.path.exists(temporary_directory):
+                        os.mkdir(temporary_directory)
+            with open(file_source, "r") as input_file, \
+                    open(file_destination, "w") as output_file:
```
Grayder0152, line 21: Use `()` instead of `\`

app/main.py
```diff
@@ -1 +1,23 @@
...
+                file_destination += file_source.split("/")[-1]
+            destination_dirs = file_destination.split("/")
+            if len(destination_dirs) > 1:
+                temporary_directory = ""
+                for folder_name in destination_dirs[:-1]:
+                    temporary_directory = os.path.join(
+                        temporary_directory, folder_name
+                    )
+                    if not os.path.exists(temporary_directory):
+                        os.mkdir(temporary_directory)
```
Grayder0152, line 19: use `makedirs` with `exist_ok=True` instead of this code

Grayder0152 (changes requested):

iishchenko:
Changes:
1)Use () instead of \
2)use makedirs with exist_ok=True instead of this code

Grayder0152 (approved):

PR #251, opened 2023-03-29
TRY

MaksNochvai:


app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if len(command.split()) == 3:
+        command_id, file_name, file_path = command.split()
```
olena-hul, line 6: Move command.split() to a variable

app/main.py
```diff
@@ -1 +1,19 @@
...
+        if command_id != "mv":
+            raise ValueError("Command should be 'mv'")
+
+        if "/" not in command:
```
olena-hul, line 10: Use `os.path.split`. "/" is a directory separator for Linux&MacOS, for Windows we use `\`

app/main.py
```diff
@@ -1 +1,19 @@
...
+            os.rename(file_name, file_path)
+        else:
+            os.makedirs(file_path, exist_ok=True)
+            os.replace(file_name, file_path)
+
+            with (open(f"{file_name}", "r") as file_in,
+                  open(f"{file_path}", "w") as file_out):
+                file_out.write(file_in.read())
+            os.remove(file_name)
```
olena-hul, line 19: `os.replace` does all the job, doesn't it? If so, you don't need lines 16-19

olena-hul (changes requested):

Bohdan-Salamakha (approved):

PR #252, opened 2023-03-31
Solution

AlexandrZhezheria:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if command.count(" ") == 2:
+        command1, source_path, destination_path = command.split()
```
ArsenPidhoretskyi, line 6:
```suggestion
        command, source_path, destination_path = command.split()
```

app/main.py
```diff
@@ -1 +1,17 @@
...
+        if "/" not in command:
+            os.rename(source_path, destination_path)
+        else:
```
ArsenPidhoretskyi, line 9: No need this case. Just do all in one.

app/main.py
```diff
@@ -1 +1,17 @@
...
+            head_path, tail = os.path.split(destination_path)
+            os.makedirs(head_path, exist_ok=True)
+            with (
+                open(f"{source_path}", "r") as file_1,
```
ArsenPidhoretskyi, line 13: rename because it is not clear names:
file_1 -> source
file_2 -> destination

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,20 @@
...
+
+def move_file(command: str) -> None:
+    if len(command.split()) == 3:
+        command, source, destination = command.split()
+        if command == "mv" and source != destination:
+            if os.path.dirname(destination):
+                try:
+                    os.makedirs(os.path.dirname(destination))
+                except FileExistsError:
+                    print("file path already created")
```
Grayder0152, line 12: use `os.path.makedirs` with params `exist_ok=True
`

Grayder0152 (changes requested):

vasylhnatiuk (approved):
Good job!

PR #263, opened 2023-04-15
Solution

bezhevets:


vasylhnatiuk (changes requested):
It is working, but you must use functions that are written in the task description, like os.mkdir, os.remove(), and others.

vasylhnatiuk (approved):
Good job!

PR #266, opened 2023-04-19
Trouble

Maksym-Filipchuk:
Two tests didnt pass 
How I can Fix it? 
have no idea (

FAILED tests/test_main.py::test_file_renamed - FileNotFoundError: [WinError 3] Системе не удается найти указанный путь: ''
FAILED tests/test_main.py::test_should_work_when_directory_exists - FileExistsError: [WinError 183] Невозможно создать файл, так как он уже существует: 'dir'

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,18 @@
-# write your code here
+from os import remove, makedirs
+
+
+def move_file(command: str):
+    source = command.split(" ")[1]
+    path_list = command.split(" ")[2]
+    path = "/".join(path_list.split("/")[0:-1])
+    with open(source, "r") as original:
+        content = original.read()
+    makedirs(path)
+    with open(path_list, "w") as new_file:
+        new_file.write(content)
+    remove(source)
+
+
+
+
+


ArsenPidhoretskyi:
Please fix flake8

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from os import remove, makedirs
+
+
+def move_file(command: str):
+    source = command.split(" ")[1]
```
ArsenPidhoretskyi, line 5: first of all you must validate that command has 3 arguments

ArsenPidhoretskyi: Also " " is redundant, because it as default
You can do command.split()

app/main.py
```diff
@@ -1 +1,18 @@
...
+    path_list = command.split(" ")[2]
+    path = "/".join(path_list.split("/")[0:-1])
```
spliting by slash is not works in all OS

ArsenPidhoretskyi (changes requested):

PR #269, opened 2023-05-01
'Solution'

kucherenkodmytro:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    mv_command, first_file, second_file_with_path = command.split()
```
vasylhnatiuk, line 5: This line must be after ` len(command.split()) == 3:` because if len is less than 3, the code will be broken here.

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):
Well done!

PR #272, opened 2023-05-06
Solution

OlehOryshchuk:


app/main.py
```diff
@@ -1 +1,26 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    content = command.split()
+    mv, move_from, file_path = content
```
olena-hul, line 6: Unpacking should be performed after the command length check

app/main.py
```diff
@@ -1 +1,26 @@
...
+    *directory, move_to = file_path.split("/")
```
olena-hul, line 7: Use os.path.split for this purpose

app/main.py
```diff
@@ -1 +1,26 @@
...
+        return
+    if move_from == move_to or move_from == file_path:
+        print("Cannot create a file when that file already exists")
+        return
+
+    dirs_path = ""
+    for dr in directory:
+        dirs_path = os.path.join(dirs_path, dr)
+        try:
+            os.mkdir(dirs_path)
```
olena-hul, line 20: You can use os.makedirs(exist_ok=True) and don't catch the exception as it will be handled automatically

app/main.py
```diff
@@ -1 +1,26 @@
...
+        except FileExistsError:
+            pass
+
+    with open(move_from, "r") as _from, open(file_path, "w") as _to:
```
olena-hul, line 24: Use better names for variables

olena-hul (changes requested):

Bohdan-Salamakha (approved):

PR #275, opened 2023-05-08
'Solution'

contritum:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    cmd, origin_file, path = command.split()
```
olena-hul, line 5: Validate command length before unpacking

app/main.py
```diff
@@ -1 +1,18 @@
...
+    if cmd == "mv":
+        if "/" in path:
+            part_of_path = path.split("/")
+            part_of_path.pop()
+            new_path = "/".join(part_of_path)
```
olena-hul, line 10: Do not use "/" to check if directory is present, it won't work on every OS. Use os.path.split and os.path.join

app/main.py
```diff
@@ -1 +1,18 @@
...
+            os.makedirs(new_path, exist_ok=True)
+            with open(origin_file, "r") as f:
+                date = f.read()
```
olena-hul, line 13: date is a bad name, mayble `data` was meant?

olena-hul (changes requested):

Bohdan-Salamakha (changes requested):
v1

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if "mv" in command:
```
ArsenPidhoretskyi, line 5: You must split command to part.
Because for example:
command = "hellomv" will return true

app/main.py
```diff
@@ -1 +1,15 @@
...
+        cmd, origin_file, path = command.split()
+        if "/" in path:
```

contritum: but command os.path.split don`t work with simple strings and I used this in the lower string with the path and folder name

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    cmd, origin_file, path = command.split()
+    if cmd == "mv":
+        if "/" in path:
```
YukihiroSM, line 7: Could you think about solution which will not use "/", because in this case the script will be not multi-platform (for example, Windows users use `\`).

YukihiroSM (changes requested):
Also, please, take a look into one of previous comments: https://github.com/mate-academy/py-move-file/pull/275/files#r1188267584

app/main.py
Bohdan-Salamakha: not fixed

Bohdan-Salamakha (changes requested):
v2

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_list = command.split()
+    if len(command_list) == 3:
+        cmd, origin_file, path = \
+            command_list[0], command_list[1], command_list[2]
```
Bohdan-Salamakha, line 8:
```suggestion
        cmd, origin_file, path = command_list
```

app/main.py
```diff
@@ -1 +1,18 @@
...
+        if cmd == "mv":
+            folders, file = os.path.split(path)
+            if len(folders) > 0:
+                os.makedirs(folders, exist_ok=True)
+                with open(origin_file, "r") as f, open(path, "w") as e:
```
Bohdan-Salamakha, line 13: use declarative names

app/main.py
```diff
@@ -1 +1,18 @@
...
+                    data = f.read()
+                    e.write(data)
```
Bohdan-Salamakha, line 15: unite these two lines

Bohdan-Salamakha (changes requested):

PR #277, opened 2023-05-10
Solution

vladimir-byshevets:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+import shutil
```
olena-hul, line 2: Don't use extra libraries. There is no profit here of using it, you can do exactly the same using os module

app/main.py
```diff
@@ -1 +1,17 @@
...
+
+
+def move_file(command: str) -> None:
+    lst = command.split(' ')
```
olena-hul, line 6:
```suggestion
    lst = command.split()
```
Space is the default argument for split

app/main.py
```diff
@@ -1 +1,17 @@
...
+    if os.path.isfile(lst[1]):
```
olena-hul, line 7: Validate command length, make sure that there are exactly 3 arguments here

app/main.py
```diff
@@ -1 +1,17 @@
...
+        if lst[2].find("/") > -1 and os.path.exists(lst[2]) is False:
```
olena-hul, line 8: Use unpacking (ex. a, b = [1, 2]) and avoid accessing command items by index. 

`lst` is a bad name for the variable

olena-hul (changes requested):

app/main.py
vladimir-byshevets: Уточните почему стоит избегать доступа к элементам команды по индексу?

olena-hul: When the list is small (2-3 items) and list items are reused several times, it's simpler to assign those items to the variables and use variables instead of hardcoding list indexes
It looks more readable

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    com_spl = command.split()
+    cmd, file_1, path_file = com_spl
+    if len(com_spl) != 3:
+        return
+    if path_file.find("/") > -1 and os.path.exists(path_file) is False:
```
olena-hul, line 9: Do not use path_file.find("/"). You can use os.path.split instead

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    com_spl = command.split()
+    cmd, file_1, path_file = com_spl
```
olena-hul, line 6: Unpacking should be done after length check

And use better names for variables

olena-hul (changes requested):

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    com_spl = command.split()
+    if len(com_spl) != 3:
+        return
+    cmd, file_old, path_file = com_spl
+    if path_file.find("/") > -1 and os.path.exists(path_file) is False:
```
ArsenPidhoretskyi, line 9: use os.path.split
instead of checking find("/")
For now, your code is not cross-platform
Did you read the checklist???
https://github.com/vladimir-byshevets/py-move-file/blob/develop/checklist.md

ArsenPidhoretskyi (changes requested):

app/main.py
vladimir-byshevets: os.path.split() Метод в Python используется для разделения имени пути на пару голов и хвостов, find - используется для поиска символа, объясните в чём ошибка?

Bohdan-Salamakha (approved):

PR #278, opened 2023-05-11
Solution

antonpiddubnyi:


app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_split = command.split()
+    if len(command_split) != 3 or "mv" not in command_split[0]:
```
ArsenPidhoretskyi, line 6: "mv" in "pleasemv" will return true
"mv" == "pleasemv" will return false
Check difference about == and in
```suggestion
    if len(command_split) != 3 or "mv" != command_split[0]:
```

app/main.py
```diff
@@ -1 +1,19 @@
...
+        return
+    cmd, file_old, path_file = command_split
+
+    if "/" in path_file:
+        path, file_name = os.path.split(path_file)
+        os.makedirs(path, exist_ok=True)
```
ArsenPidhoretskyi, line 12: It is redundant code
As for me it is not needed
Also check for / is not good way because it will not work in all OS

ArsenPidhoretskyi (changes requested):

app/main.py
antonpiddubnyi: @ArsenPidhoretskyi so what should it be changed to? How do I check that there are multiple levels of directory?

ArsenPidhoretskyi (approved):
Well done!

PR #279, opened 2023-05-11
Solution

VladyslavKaminskyi:


app/main.py
```diff
@@ -1 +1,14 @@
...
+from os import makedirs, path, remove
+
+
+def move_file(command: str) -> None:
+    command_list = command.split()
+    if len(command.split()) != 3 or command_list[0] != "mv":
+        raise Exception("Wrong command!")
+    mv, incoming_file, new_file = command_list
+    if "/" in new_file:
+        makedirs(path.dirname(new_file), exist_ok=True)
```
ArsenPidhoretskyi, line 10: Better to do like this, because it will not work in some os (check for "/"
```suggestion
    dir_path, file_name = path.split(new_file)
    if dir_path:
        makedirs(dir_path, exist_ok=True)
```

app/main.py
```diff
@@ -1 +1,14 @@
...
+    with (open(incoming_file, "r") as incoming,
+          open(new_file, "w") as new):
```
ArsenPidhoretskyi, line 12:
```suggestion
    with (
        open(incoming_file, "r") as incoming,
        open(new_file, "w") as new
    ):
```

ArsenPidhoretskyi (changes requested):

PR #280, opened 2023-05-12
crazy tests

sbilikepy:


app/main.py
```diff
@@ -1 +1,33 @@
...
+    if "/" not in rename_case_flow[2]:
+        os.rename(rename_case_flow[1], rename_case_flow[2])
+
+    else:
+        source_name = move_case_flow[0].split(" ")[1]
+        new_name = move_case_flow[-1]
+        path = [f"{move_case_flow[0].split(' ')[2]}"]
+        mkdir_path = ""
+
+        for i in move_case_flow[1:-1]:
```
Vasyl-Smutok, line 20: `i` bad variable name

app/main.py
```diff
@@ -1 +1,33 @@
...
+            path.append(i)
+        for i in range(len(path)):
+            try:
+                mkdir_path += f"{path[i]}/"
+                os.mkdir(mkdir_path)
```
Vasyl-Smutok, line 25: try to use `os.makedirs`

app/main.py
```diff
@@ -1 +1,33 @@
...
+            except Exception as e:
+                print(f"{e}")
```
Vasyl-Smutok, line 27: add more information of Exception

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+        with (open(f"{source_name}", "r") as source_file,
+              open(f"{mkdir_path + new_name}", "w") as file_copy):
+            content = source_file.read()
+            file_copy.write(content)
```
Vasyl-Smutok, line 32:
```suggestion
            file_copy.write(source_file.read())
```

Vasyl-Smutok, line 32:
```suggestion
            file_copy.write(source_file.read())
```

vsmutok (changes requested):
also remove `file1.txt` from PR

vsmutok (changes requested):
also remove `file1.txt` from PR

app/main.py
```diff
@@ -1 +1,39 @@
-# write your code here
+import os
+
+
+class PlannedError(Exception):
+    """To create os.mkdir("first/second"), first directory needs to exist."""
```
Vasyl-Smutok, line 5: why do you need this custom exception?
if you don't use anywhere?

app/main.py
```diff
@@ -1 +1,39 @@
...
+
+                mkdir_path += f"{path[new_folder]}/"
+                os.makedirs(mkdir_path)
+            except PlannedError:
+                print("The path is already created. "
+                      "os.makedirs failed as the root directory "
+                      "already exists.")
+
+        with (open(f"{source_name}", "r") as source_file,
+              open(f"{mkdir_path + new_name}", "w") as file_copy):
```
Vasyl-Smutok, line 37:
```suggestion
    with (
        open(f"{source_name}", "r") as source_file,
        open(f"{mkdir_path + new_name}", "w") as file_copy
    ):
```

vsmutok (changes requested):

app/main.py
sbilikepy: i'll use `FileExistsError` instead of it then

app/main.py
```diff
@@ -1 +1,34 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    rename_case_flow = command.split(" ")
+    move_case_flow = command.split("/")
```
ArsenPidhoretskyi, line 6: This line is so strange. Remove it no need to split by /

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+    if rename_case_flow[0] != "mv":
+        raise ValueError("Wrong command")
+
+    if "/" not in rename_case_flow[2]:
```
ArsenPidhoretskyi, line 11: If you want check by / use os.path.split
Because in some os path will be `\\` instead of `/`

app/main.py
```diff
@@ -1 +1,34 @@
...
+        mkdir_path = ""
+
+        for path_component in move_case_flow[1:-1]:
+            path.append(path_component)
+        for new_folder in range(len(path)):
+            try:
+                mkdir_path += f"{path[new_folder]}/"
+                os.makedirs(mkdir_path)
+            except FileExistsError as e:
+                print(f"{e}")
```
ArsenPidhoretskyi, line 27: What do you do here? I do not understand this logic totally

app/main.py
```diff
@@ -1 +1,34 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    rename_case_flow = command.split(" ")
+    move_case_flow = command.split("/")
+
+    if rename_case_flow[0] != "mv":
```
ArsenPidhoretskyi, line 8: Also validate that command.split() have 3 parameters

app/main.py
```diff
@@ -1 +1,34 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    rename_case_flow = command.split(" ")
```
ArsenPidhoretskyi, line 5: rename this variable.
image

Rename it to command_parameters

app/main.py
ArsenPidhoretskyi: you can do:
cmd, source_path, destination_path = command_split

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    cmd, source_path, destination_path = command.split(" ")
```
ArsenPidhoretskyi, line 5: Can you check:
command == "mv"
It it will raise an error? As for me yes. You need validate for command.split() before this line

app/main.py
```diff
@@ -1 +1,18 @@
...
+
+    if len(command.split()) == 3 and "mv" in cmd:
```
ArsenPidhoretskyi, line 7: when cmd will "hello_mv" be it will return true. Check difference between in and ==

app/main.py
```diff
@@ -1 +1,18 @@
...
+
+        if "/" not in destination_path:  # because name cannot contain "/"
+            os.rename(source_path, destination_path)
+
+        os.makedirs(os.path.split(destination_path)[0])
```
ArsenPidhoretskyi, line 12: As for me this code is not needed. Can you check?
If needed checking for / is so bad. Because in some OS it will not work. Read about os.path.split

app/main.py
```diff
@@ -1 +1,18 @@
...
+import os
+
+
+def move_file(command: str) -> None:
+    cmd, source_path, destination_path = command.split(" ")
+
+    if len(command.split()) == 3 and "mv" in cmd:
+
+        if "/" not in destination_path:  # because name cannot contain "/"
+            os.rename(source_path, destination_path)
```
ArsenPidhoretskyi, line 10: it this case you also must do return

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (changes requested):
Changes requested

app/main.py
sbilikepy: > Can you check: command == "mv" It it will raise an error? As for me yes. You need validate for command.split() before this line
should i try:
```
    if isinstance(command, str) and len(command.split()) == 3:
        cmd, source_path, destination_path = command.split()
```
?

app/main.py
sbilikepy: > As for me this code is not needed. Can you check? If needed checking for / is so bad. Because in some OS it will not work. Read about os.path.split

for `move_file("mv rename.txt rename2.txt")` result will be: 
['mv', 'rename.txt', 'rename2.txt']
('', 'mv rename.txt rename2.txt')

and for `move_file("mv move.txt first_dir/second_dir/move2.txt")`:
['mv', 'move.txt', 'first_dir/second_dir/move2.txt']
('mv move.txt first_dir/second_dir', 'move2.txt')

i can replace "/" check by `if os.path.split(command)[0] == ""`

sbilikepy:
> os.makedirs

sorry i didn't get how can i skip `os.makedirs` here

app/main.py
```diff
@@ -1 +1,23 @@
...
+        if cmd == "mv":
+
+            if os.path.split(command)[0] == "":
+                try:
+                    os.rename(source_path, destination_path)
+                except FileNotFoundError:
+                    print(f"{source_path} not found")
+
+            else:
+                os.makedirs(os.path.split(destination_path)[0])
```
Grayder0152, line 17:
```suggestion
                os.makedirs(os.path.split(destination_path)[0], exist_ok=True)
```

Grayder0152 (changes requested):

vasylhnatiuk (approved):
Good job!

PR #283, opened 2023-05-16
Soution

MykytaKuzmytskyi:


tests/test_main.py
```diff
@@ -17,13 +17,13 @@ def create_file(filename: str = "file.txt") -> str:
...
-    move_file("mv file.txt file1.txt")
+    move_file("mv file.txt file2.txt")
 
     assert os.path.exists("file.txt") is False
-    with open("file1.txt", "r") as file_with_content:
+    with open("file2.txt", "r") as file_with_content:
         assert file_with_content.read() == "This is some\n content for\n the file."
 
-    os.remove("file1.txt")
+    os.remove("file2.txt")
```
Bohdan-Salamakha, line 26: don't change tests

Bohdan-Salamakha (changes requested):
v1

app/main.py
```diff
@@ -1 +1,38 @@
-# write your code here
+import os
+import shutil
+
+
+def move_file(command: str) -> None:
+    split_command = command.split()
+    exist_file, new_name = split_command[1], split_command[2]
```
olena-hul, line 7: Use unpacking, no need to access commands by indexes

app/main.py
```diff
@@ -1 +1,38 @@
...
+
+    directories = split_command[-1].split("/")
```
olena-hul, line 9: Don't split by "/" symbol. Use os.path.split

app/main.py
```diff
@@ -1 +1,38 @@
...
+
+    if len(directories) == 1:
+        return os.rename(exist_file, new_name)
+
+    dst_file1 = parent_dir + "\\" + directories[-2] + "\\" + exist_file
+    dst_file2 = parent_dir + "\\" + directories[-2] + "\\" + new_file
+
+    path = os.getcwd()
+    for directorie in directories[:-1]:
+        path += "\\" + directorie
```
olena-hul, line 24: Your code should be cross-platformed. This will work only on windows. Rewrite it by using neccessary functions from os module, such as os.path.split and os.path.join

olena-hul (changes requested):

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+def move_file(command: str) -> None:
+    split_command = command.split()
+    comm, exist_file, path = split_command
+    path, new_file = os.path.split(path)
+
+    src_file = os.path.join(os.getcwd(), exist_file)
+
+    if not path:
+        return os.rename(exist_file, new_file)
```
YukihiroSM, line 13: you shouldn't return the function value. Now it returns None -> that's ok. But what if tomorrow developers will change this to return some value? You should separate this into lines. 
Same for `return os.rename(dst_file1, dst_file2)`

YukihiroSM (changes requested):

YukihiroSM (approved):
Good job.

PR #287, opened 2023-05-18
py-move-file-v1

muriakula:


app/main.py
```diff
@@ -1 +1,17 @@
...
+    elements = command.split()
+    if len(elements) == 3:
+        command, file_in, file_path = elements
+        if command == "mv":
+            path_elements = file_path.split("/")
+            path = path_elements[0]
+            for element in path_elements[1:]:
+                if not os.path.isdir(path):
+                    os.mkdir(path)
+                path += "/" + element
```
YukihiroSM, line 14: it's better to use os.path.join in this case.

app/main.py
```diff
@@ -1 +1,17 @@
...
+
+
+def move_file(command: str) -> None:
+    elements = command.split()
+    if len(elements) == 3:
+        command, file_in, file_path = elements
+        if command == "mv":
+            path_elements = file_path.split("/")
+            path = path_elements[0]
+            for element in path_elements[1:]:
```
YukihiroSM, line 11: Also, try to do this without using loop. There is a function in os lib for creating several dirs.

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    elements = command.split()
+    if len(elements) == 3:
+        command, file_in, file_path = elements
+        if command == "mv":
+            path_elements = file_path.split("/")
```
YukihiroSM, line 9: what if the user uses windows, where the path is being split by `\`

YukihiroSM (changes requested):

PR #288, opened 2023-05-19
Implemented `move_file` function

N1khto:


app/main.py
```diff
@@ -1 +1,27 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if len(command.split(" ")) != 3 or command.split(" ")[0] != "mv":
+        return
+    file_orig, file_goal = command.split(" ")[1:]
```
ArsenPidhoretskyi, line 7: Splitting 3 times is so bad....... Also check for the same files is not exist
```suggestion
    command_args = command.split()
    if len(command_args) != 3:
        return
    
    cmd, file_orig, file_goal = command_args
```

app/main.py
```diff
@@ -1 +1,27 @@
...
+    if "/" not in file_goal:
```
ArsenPidhoretskyi, line 8: use os.path.join instead of checks by /

ArsenPidhoretskyi (changes requested):
Code does not work in all OS. Use os.path for split and join by / or \\ based on OS(it works auto)

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_args = command.split()
+    if len(command_args) != 3 or command_args[0] != "mv":
+        return
```
Vasyl-Smutok, line 7: I think better raise Exception here

Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1 +1,18 @@
...
+    command_args = command.split()
+    if len(command_args) != 3 or command_args[0] != "mv":
+        raise Exception("Invalid command")
+    cmd, file_orig, file_goal = command_args
+    path, file = os.path.split(file_goal)
+    if not path:
+        os.rename(file_orig, file_goal)
+        return
+    if not file:
+        file_goal = os.path.join(file_goal, file_orig)
```
ArsenPidhoretskyi, line 14: Nothing said in task that we need to do that

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #289, opened 2023-05-20
Implement move file task

vitalii-babiienko:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_name, source, destination = command.split()
```
Vasyl-Smutok, line 5: here we will get an error if the wrong command is passed, so first check if you can split it into 3 parts

Vasyl-Smutok (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #291, opened 2023-05-23
Solution

OksanaAdamchuk:


app/main.py
```diff
@@ -1 +1,32 @@
...
+
+def move_file(command: str) -> None:
+    commands = command.split()
+    if len(commands) != 3:
+        raise ValueError("The command should contain the name of the command, "
+                         "source file name, and directory to move")
+    elif commands[0] != "mv":
+        raise ValueError("The command should begin from 'mv'")
+    elif commands[2].find("/") == -1:
+        os.rename(commands[1], commands[2])
```
ArsenPidhoretskyi, line 12: Not true. In some os path looks like '\' or '\\'
Use os.path.split

ArsenPidhoretskyi: Also the rest code have problem with OS compability

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #297, opened 2023-05-27
Solution

vlamuha:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    parts_of_command = command.split()
+
+    if parts_of_command[0] != "mv" or len(parts_of_command) != 3:
```
ArsenPidhoretskyi, line 7: Change order of checks because if len of command will be 0 it will throw error in this case
```suggestion
    if len(parts_of_command) != 3 or parts_of_command[0] != "mv":
```

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

Abnormaltype (approved):

PR #299, opened 2023-05-29
Solution

Nagerba:


app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    try:
+        cmd, file, path = command.split()
```
olena-hul, line 6: Use full names, better to name the variable `command` instead of `cmd`

app/main.py
```diff
@@ -1 +1,22 @@
...
+    except ValueError:
+        print("Command should have 2 arguments")
+        return
+
+    if cmd != "mv":
+        return
+
+    dirs = "/".join(path.split("/")[:-1])
```
olena-hul, line 14: Don't use `/` to work with paths. Use functions from os module: os.path.join and os.path.split

app/main.py
```diff
@@ -1 +1,22 @@
...
+
+    if dirs:
+        os.makedirs(dirs)
```
olena-hul, line 17: This will raise an exception if the directories already exist

olena-hul (changes requested):

EdAlekseiev (approved):
Good job

PR #300, opened 2023-05-29
'Solution'

Jorjio22:


app/main.py
```diff
@@ -1 +1,53 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    # Exclude input without the command "mv"
+    if command.split()[0] != "mv":
```
EdAlekseiev, line 6: Don't use `split()` twice. Create a variable that will store the result of `split()`

app/main.py
```diff
@@ -1 +1,53 @@
...
+        print("The function only supports the 'mv' command!")
+        return
+
+    directories = command.split("/")
```
EdAlekseiev, line 10: Don't use `split("/")` for working with paths. Use os.path.split.

app/main.py
```diff
@@ -1 +1,53 @@
...
+    ]
+    first_file_name = directories[0].split()[1]
+    second_file_name = directories[-1]
+    directory_path = os.path.join("/".join(directory_names))
+    file_path = os.path.join(directory_path, second_file_name)
+
+    # If the original file needs to be copied to the same directory, rename it.
+    if len(directories) == 1:
+        second_file_name = directories[0].split()[-1]
+        os.rename(first_file_name, second_file_name)
```
EdAlekseiev, line 29: You do a lot of things. You need split `command`. The second element of the result will be a path to the source file. The third element will be the path to new_file, you must split it by using `os.path.split`. Create dirs with using `os.mkdir` and write content to new file from source file

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    # Exclude the creation of a directory without a file.
+    if command.endswith("/"):
+        print("Specify the file name after '/'!")
+        return
+
+    instruction = command.split()
+    first_file_name = instruction[1]
+    path = instruction[-1].split("/")[0:-1]
+    source_path, second_file_name = os.path.split(command.split()[-1])
```
YukihiroSM, line 18:
```suggestion
    source_path, second_file_name = os.path.split(instruction[-1])
```

app/main.py
```diff
@@ -1 +1,48 @@
...
+        return
+
+    # Exclude the creation of a directory without a file.
+    if command.endswith("/"):
+        print("Specify the file name after '/'!")
+        return
+
+    instruction = command.split()
+    first_file_name = instruction[1]
+    path = instruction[-1].split("/")[0:-1]
```
YukihiroSM, line 17: use os.path.split, please.

YukihiroSM (changes requested):

PR #301, opened 2023-05-29
Solution

Munkoor:


app/main.py
```diff
@@ -1 +1,20 @@
...
+
+
+def move_file(command: str) -> None:
+    split_command = command.split()
+
+    if len(split_command) == 3:
+        comm, file_in, file_path = split_command
+
+        if comm == "mv":
+            path = os.path.dirname(file_path)
```
ArsenPidhoretskyi, line 11: Why you need dirname?
YOu can use os.path.split

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,22 @@
...
+        comm, file_in, file_path = split_command
+
+        if comm == "mv":
+            path, _ = os.path.split(file_path)
+            if not path:
+                path = "./"
+            if file_path.endswith("/"):
+                path = file_path
+            elif len(path) > 1 and not os.path.exists(path):
+                os.makedirs(path, exist_ok=True)
```
ArsenPidhoretskyi, line 17: So hard logic as for me. Simplify it and just do:
```suggestion
            if path:
                 os.makedirs(path, exist_ok=True)
```

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    split_command = command.split()
+
+    if len(split_command) == 3:
+        comm, file_in, file_path = split_command
```
vasylhnatiuk, line 8: comm is a bad name

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):
Good job

PR #303, opened 2023-05-31
Solution

gleblitvinenko:


app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+import os
+
+
+def move_file(commnad: str) -> None:
```
vsmutok, line 4: Typo in the function argument name 'commnad', it should be 'command'.

app/main.py
```diff
@@ -1 +1,19 @@
...
+    try:
+        command, current_file_name, new_file_path = commnad.split()
+    except ValueError:
+        print("Command must have 3 arguments")
+        return
+    if current_file_name == new_file_path or command != "mv":
```
vsmutok, line 10: It's better to use a separate condition to check if the command is 'mv' and print an error message if it's not.

app/main.py
```diff
@@ -1 +1,19 @@
...
+        return
+    destination_path = os.path.dirname(new_file_path)
```
vsmutok, line 12: Why you need `dirname`?
You can use `os.path.split`

vsmutok (changes requested):

Grayder0152 (approved):

PR #308, opened 2023-06-06
mv implemented

thorium-zero:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+from shutil import move
+from os import makedirs
+from os.path import dirname
+
+
+def move_file(command: str) -> None:
+    params = command.split()
```
YuriiKindrat, line 7: try to unpack this like this `arg1,arg2,arg3 = command.split() `

YuriiKindrat (changes requested):

app/main.py
thorium-zero: Again, I'll have to handle possible exception here if I do that

YuriiKindrat: agree, but u can unpack this after ur condition 
arg1, arg2, arg3 = params
it seems to me, that for better readability is is better to use vars instead of indexes

YuriiKindrat: but it is up to u

YuriiKindrat (changes requested):

app/main.py
thorium-zero: Yeah, that was the solution I was proposing in the previous ("copy file") project you've rejected as my compromise.

app/main.py
```diff
@@ -1 +1,12 @@
...
+    if len(params) == 3 and params[0] == "mv":
+        dest_dir = dirname(params[2])
+        if dest_dir != "":
```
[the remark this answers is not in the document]
thorium-zero: Yes, I know. I simply dislike this approach. And I don't like it that python accepts implicit bool conversions with conditionals just as c/c++ does. This approach is potentially dangerous and should be avoided when possible. Although, it is python we are dealing with here... So it is, probably, not so big of an issue considering the nature of the language itself... 🤣

Oleg-s-s (approved):
Looks great :+1:

PR #311, opened 2023-06-08
Develop

n1ckmel:


vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_parts = command.split(" ")
+    if command_parts[0] != "mv" or len(command_parts) != 3:
```
ArsenPidhoretskyi, line 6: Change order because if len(command_parts) is 0 it will give exception
```suggestion
    if len(command_parts) != 3 or command_parts[0] != "mv":
```

app/main.py
```diff
@@ -1 +1,20 @@
...
+        raise ValueError("Command is incorrect")
+    file_name = command_parts[1]
+    new_file_path = command_parts[2]
```
ArsenPidhoretskyi, line 9:
```suggestion
    file_name, new_file_path = command_parts[1:]
```

app/main.py
```diff
@@ -1 +1,20 @@
...
+    dir_list = command_parts[2].split("/")[:-1]
```
ArsenPidhoretskyi, line 10: Use os.path.split

app/main.py
```diff
@@ -1 +1,20 @@
...
+    path = ""
+    for directory in dir_list:
+        path = os.path.join(path, directory)
+        if not os.path.exists(path):
+            os.mkdir(path)
```
ArsenPidhoretskyi, line 15: Use os.makedirs with exist_ok=True

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #312, opened 2023-06-08
Solution

ShevchukVitalii:


app/main.py
```diff
@@ -1 +1,21 @@
...
+import os
+
+
+def move_file(command: str) -> None:
+    parts = command.split()
+
+    if len(parts) != 3:
+        raise ValueError("Invalid command")
+
+    if parts[0] != "mv":
```
ArsenPidhoretskyi, line 10:
```suggestion
    cmd, source_path, destination_path = parts
    if cmd != "mv":
```
And the use this variables

ArsenPidhoretskyi: Unzipiing better than [0], [1]....

ArsenPidhoretskyi (changes requested):

app/main.py
ShevchukVitalii: agree with you

PR #313, opened 2023-06-09
Solution

diana-shyrokikh:


app/main.py
```diff
@@ -1 +1,48 @@
...
+            "the command should look like this: "
+            "mv file.txt some_dir/new_file.txt"
+        )
+
+    file_to_move = command[1]
+
+    if not path.exists(file_to_move):
+        raise FileNotFoundError(f"File {file_to_move} does not exist")
+
+    path_of_new_file = command[-1].split("/")
```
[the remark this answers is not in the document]
AntonShpakovych: and after this u can check if this dir already presented u do moving if not u create this dir
`if not os.path.exists(dir_path): os.makedirs(dir_path)`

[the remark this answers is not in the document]
AntonShpakovych: And also think about module `shutil` which have method move

AntonShpakovych (changes requested):

app/main.py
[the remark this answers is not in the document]
AntonShpakovych: And think about `PermisionError`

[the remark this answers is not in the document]
AntonShpakovych: You don't need to specify "r" - mode he set by default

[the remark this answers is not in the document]
diana-shyrokikh: Thanks, I hope I understood you correctly

app/main.py
```diff
@@ -1 +1,45 @@
...
+
+    if path.exists(command[-1]):
+        raise PermissionError("The file cannot be overwritten")
+
+    file_to_move = command[1]
+
+    if not path.exists(file_to_move):
+        raise FileNotFoundError(f"File {file_to_move} does not exist")
+
+    is_consist_directory = True if "/" in command[-1] else False
```
AntonShpakovych, line 35: You don't need to use ternary operation here. U can can rewrite this to `/` in command[-1]

app/main.py
```diff
@@ -1 +1,45 @@
...
+
+    if len(command) != 3:
+        raise MoveCommandSyntaxError(
+            "Invalid move command syntax.\n"
+            "For a successful moving of file "
+            "the command should look like this: "
+            "mv file.txt some_dir/new_file.txt"
+        )
+
+    if path.exists(command[-1]):
```
AntonShpakovych, line 27: PermissionError u need to use in try except where u read your file, because user can give you a file which u can't read, u don't have permission

AntonShpakovych (changes requested):
So fixed, permission error and ternary operation and I will approve.  Like in your logic u raise Permission when file already exist, but this file can be empty. And u don't validate when user give u a file which u can't read, because you don't have permission for this.

app/main.py
```diff
@@ -1 +1,48 @@
...
+            "the command should look like this: "
+            "mv file.txt some_dir/new_file.txt"
+        )
+
+    file_to_move = command[1]
+
+    try:
+        with open(file_to_move) as source:
+            source.read()
+    except PermissionError:
```
AntonShpakovych, line 32: okey) But here u already check permission but if u want read a file which doesn't exist

AntonShpakovych (changes requested):

app/main.py
```diff
@@ -1 +1,48 @@
...
+        raise MoveCommandSyntaxError(
+            "Invalid move command syntax.\n"
+            "For a successful moving of file "
+            "the command should look like this: "
+            "mv file.txt some_dir/new_file.txt"
+        )
+
+    file_to_move = command[1]
+
+    if not path.exists(file_to_move):
```
AntonShpakovych, line 29: why u check it here u can do it like with Permission error

app/main.py
```diff
@@ -1 +1,48 @@
...
+        raise FileNotFoundError(f"File {file_to_move} does not exist")
+
+    try:
+        with open(file_to_move) as source:
+            source.read()
+    except PermissionError:
+        raise PermissionError(f"No permission to read the file {source}")
```
AntonShpakovych, line 36: Why here u use message? Like default message from `PermissionError` is  the same, why in except u raise `PermissionError` manually? think about correct re-rasing

AntonShpakovych (changes requested):

app/main.py
AntonShpakovych: We have info about correct re-rasing on mate platform

app/main.py
```diff
@@ -1 +1,48 @@
...
+        )
+
+    file_to_move = command[1]
+
+    if not path.exists(file_to_move):
+        raise FileNotFoundError(f"File {file_to_move} does not exist")
+
+    try:
+        with open(file_to_move) as source:
+            source.read()
```
AntonShpakovych, line 34: you  can avoid this. You use source.read() just for handle some errors or what?

diana-shyrokikh: yes, I try to handle error

AntonShpakovych (approved):
Good job!

app/main.py
```diff
@@ -1 +1,41 @@
...
+
+class MoveCommandSyntaxError(Exception):
+    pass
+
+
+def move_file(command: str) -> None:
+    command = command.split()
+
+    if command[0] != "mv":
+        raise MoveCommandSyntaxError("Move command not found")
```
YuriiKindrat, line 17: u cna use startswith before spliting for this )

app/main.py
```diff
@@ -1 +1,41 @@
...
+
+    if len(command) != 3:
+        raise MoveCommandSyntaxError(
+            "Invalid move command syntax.\n"
+            "For a successful moving of file "
+            "the command should look like this: "
+            "mv file.txt some_dir/new_file.txt"
+        )
```
YuriiKindrat, line 25: u can unpack ur command after this check to avoid using indexes & better readability

app/main.py
```diff
@@ -1 +1,41 @@
...
+
+    file_to_move = command[1]
+    is_consist_directory = "/" in command[-1]
+
+    if is_consist_directory:
+        path_of_new_file = path.dirname(command[-1])
+
+        if not path.exists(path_of_new_file):
+            makedirs(path_of_new_file)
```
YuriiKindrat, line 34: could u use param `exist_ok` instead of if condition?

app/main.py
```diff
@@ -1 +1,41 @@
...
+
+    path_of_new_file = command[-1]
+
+    try:
+        move(file_to_move, path_of_new_file)
+    except (FileNotFoundError, PermissionError):
+        raise
```
YuriiKindrat, line 41: why do u need to catch and then reraise an errors ?

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,37 @@
...
+    if len(command) != 3:
+        raise MoveCommandSyntaxError(
+            "Invalid move command syntax.\n"
+            "For a successful moving of file "
+            "the command should look like this: "
+            "mv file.txt some_dir/new_file.txt"
+        )
+
+    mv_command, file_to_move, new_file_path = command
+    is_consist_directory = "/" in command[-1]
```
LiudmylaKulinchenko, line 28: Is ur code cross-platform?) For example, will it work on win with \ ? Use os.path.split() instead

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! 😋🐇

PR #315, opened 2023-06-11
Solution

AntonShpakovych:
I used raise without message, because my function return bool value, so for me its not necessary

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -7,7 +7,7 @@ def move_file(command: str) -> bool:
         command, file, new_file = command.split(" ")
         consist_directory = "/" in new_file
 
-        if command != 'mv':
+        if command != "mv":
             raise ValueError
         elif consist_directory:
             dir_path = os.path.dirname(new_file)
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,25 @@
-# write your code here
+import os
+import shutil
+
+
+def move_file(command: str) -> bool:
+    try:
+        command, file, new_file = command.split(" ")
+        consist_directory = "/" in new_file
+
+        if command != 'mv':
+            raise ValueError
+        elif consist_directory:
+            dir_path = os.path.dirname(new_file)
+
+            if not os.path.exists(dir_path):
+                os.makedirs(dir_path)
+            shutil.move(file, new_file)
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,25 @@
-# write your code here
+import os
+import shutil
+
+
+def move_file(command: str) -> bool:
+    try:
+        command, file, new_file = command.split(" ")
```
YevheniiIovchu, line 7: It's not the best practice to change the value of function argument

YevheniiIovchu (changes requested):

app/main.py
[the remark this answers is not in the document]
AntonShpakovych: Yes, I know. But its depend to people. I like when I use split with " "

AntonShpakovych: Oh yes, thx

YevheniiIovchu (approved):

app/main.py
```diff
@@ -1 +1,23 @@
...
+
+
+def move_file(command: str) -> str:
+    try:
+        current_command, file, new_file = command.split()
+
+        if current_command != "mv" or file == new_file:
+            raise ValueError
+
+        consist_directory = "/" in new_file
```
YuriiKindrat, line 14: will this work with all OS ?

app/main.py
```diff
@@ -1 +1,23 @@
...
+import os
+import shutil
+
+
+def move_file(command: str) -> str:
+    try:
+        current_command, file, new_file = command.split()
+
+        if current_command != "mv" or file == new_file:
+            raise ValueError
```
YuriiKindrat: add message when u are raising exceptions

app/main.py
```diff
@@ -1 +1,23 @@
...
+
+        consist_directory = "/" in new_file
+
+        if consist_directory:
+            dir_path = os.path.dirname(new_file)
+
+            if not os.path.exists(dir_path):
+                os.makedirs(dir_path)
```
YuriiKindrat, line 20: could u use param `exist_ok` for makedirs here ?

YuriiKindrat (changes requested):

PR #316, opened 2023-06-11
solution

Tarasidze:


Kirontiko (changes requested):

app/main.py
```diff
@@ -1 +1,23 @@
...
+import os
+
+
+def move_file(command: str) -> any:
+    commands = command.split(" ")
+
+    if len(commands) == 3 and command.startswith("mv "):
+        _, first_file_name, path_to_second_file = commands
+        bare_path_to_second_file = os.path.dirname(path_to_second_file)
+        second_file_name = path_to_second_file.split("/")[-1]
```
LiudmylaKulinchenko, line 11: Is ur code cross-platform?) For example, will it work on win with \ ? Use os.path.split() instead

app/main.py
```diff
@@ -1 +1,23 @@
-# write your code here
+import shutil
+import os
+
+
+def move_file(command: str) -> any:
+    commands = command.split(" ")
```
LiudmylaKulinchenko, line 6: no need to use space inside split method, it uses space by default

LiudmylaKulinchenko (changes requested):

app/main.py
Tarasidze: If only I knew about os.path.split() yesterday, I would save a lot of time)

app/main.py
```diff
@@ -1 +1,22 @@
...
+
+def move_file(command: str) -> None:
+    commands = command.split()
+
+    if len(commands) == 3 and command.startswith("mv "):
+        _, first_file_name, path_to_second_file = commands
+        (path_to_second_file,
+            second_file_name) = os.path.split(path_to_second_file)
+
+        if path_to_second_file == "":
```
LiudmylaKulinchenko, line 13: Using `if not path_to_second_file` will make the condition more readable and general

LiudmylaKulinchenko (changes requested):

app/main.py
Tarasidze: I don't know why, but yesterday it didn't work. Today is OK.

Oleg-s-s (approved):
:+1:

PR #317, opened 2023-06-11
solution

oleksiikolii:


app/main.py
```diff
@@ -1 +1,23 @@
...
+class CommandError(Exception):
+    pass
+
+
+def move_file(command: str) -> None:
+    command = command.split()
+    if len(command) != 3:
+        raise CommandError(f"Wrong format of command: {command}")
+
+    if command[0] != "mv":
```
Tarasidze, line 13: There is a better way to write a condition: 
 - do not change the incoming string and write down: if command.startswith("mv ")

app/main.py
```diff
@@ -1 +1,23 @@
...
+        raise CommandError(f"Wrong command:"
+                           f"'{command[0]}', did you mean 'mv'?")
+
+    file_name, full_path = command[1], command[2]
```
Tarasidze, line 17: In my opinion " _, file_name, full_path = command" looks more readable.
Also, try not to use indexing (command[0]) if it is possible

Tarasidze (changes requested):

app/main.py
oleksiikolii: `command` already split in parts and I need it splitted to unpack other parts of command later in code, your suggestion will break code logic because `command` already is list object which has no `startswith ()` method and also this will add unnecessary check before splitting. Why do I need to do this if I will later split it still?

app/main.py
oleksiikolii: I don't think I need to write unnecessary `_` here, it has no impact on logic makes additional link to object that I will never use and less readable
And for no using command[0] I just don't want to create new variable to use it once

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 😋🌼

PR #320, opened 2023-06-12
My solution

kovaliskoveronika:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command = command.split()
```
volodymyr-komarnyckyi, line 5: You can make something like cmd, old_file, new_file = command to improve code readability

app/main.py
```diff
@@ -1 +1,10 @@
...
+    path = os.path.dirname(command[2])
```
volodymyr-komarnyckyi, line 6: Add some checks to confirm that the command is correct and source file exists, if not - raise error

app/main.py
volodymyr-komarnyckyi, line 5: Make it like try block and in the end add exceptions if something went wrong

app/main.py
volodymyr-komarnyckyi, line 6: Maybe rename the variable to destination_path for better understanding

volodymyr-komarnyckyi (changes requested):

app/main.py
```diff
@@ -1 +1,16 @@
...
+        print("Wrong command")
+        return
+    try:
+        destination_path = os.path.dirname(copy)
+        if destination_path:
+            if not os.path.exists(destination_path):
+                os.makedirs(destination_path)
+        os.rename(origin, copy)
+    except FileNotFoundError:
+        print("There is no file with this name")
```
Oleg-s-s, line 16: Looks like you can check if file exist using `if` statement to improve code efficiency

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -2,9 +2,15 @@
...
-    command = command.split()
-    path = os.path.dirname(command[2])
-    if path:
-        if not os.path.exists(path):
-            os.makedirs(path)
-    os.rename(command[1], command[2])
+    command, origin, copy = command.split()
+    if command != "mv":
+        print("Wrong command")
+        return
```
volodymyr-komarnyckyi, line 8: No need to use return in another line, you can return print

volodymyr-komarnyckyi (changes requested):

Oleg-s-s (approved):
Looks cool)

PR #322, opened 2023-06-12
Solution

volodymyr-komarnyckyi:


app/main.py
```diff
@@ -1 +1,24 @@
...
+
+def move_file(command: str) -> None:
+    try:
+        cmd, source, destination = command.split()
+        if cmd != "mv":
+            raise ValueError("Invalid command!")
+        if not os.path.exists(source):
+            raise FileNotFoundError(f"{source} doesn't exists!")
+        if destination.endswith("/"):
+            raise ValueError("There's no name for new file!")
```
YuriiKindrat, line 13: what if u will use different OS?
win os is using backslashes)

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #323, opened 2023-06-12
Solution for Move File

OlhaTryhub:


app/main.py
```diff
@@ -1 +1,25 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+
+    mv, old_file_name, new_path_file_name = command.split(" ")
```
YuriiKindrat, line 6: no need to use space inside split method, it uses space by default

app/main.py
```diff
@@ -1 +1,25 @@
...
+        new_file_name = old_file_name
+    else:
+        new_file_name = new_path_file_name.split("/")[-1]
+        directories = new_path_file_name.replace(new_file_name, "")
+
+    try:
+        if directories != "":
+            os.makedirs(directories)
+    except FileExistsError:
+        pass
```
YuriiKindrat, line 19: u can use exist_ok param for makedirs here)

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,23 @@
...
+import os
+
+
+def move_file(command: str) -> None:
+
+    mv, old_file_name, new_path_file_name = command.split()
+
+    path = os.path.dirname(new_path_file_name)
+
+    new_file_name = new_path_file_name.split("/")[-1]
```
LiudmylaKulinchenko, line 10: Use  os.path.split instead of splitting by /. It will make your code cross-platform
Also remember about os.path.join

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🐇

PR #326, opened 2023-06-16
'Solution'

sind14:


Bohdan-Salamakha (changes requested):
fix flake8

Bohdan-Salamakha (changes requested):
use os.makedirs(dest_dir, exist_ok=True) to simplify the logic

Bohdan-Salamakha (approved):

PR #327, opened 2023-06-16
Solution

drivemax26:


app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+import os
+
+
+def move_file(command):
+    command = command.split()
```
YukihiroSM, line 5: don't change the input objects. Use another variable name.

app/main.py
```diff
@@ -1 +1,14 @@
...
+    if "/" not in command[2]:
```
YukihiroSM, line 6: Check for length of command (it may contain 2 elements instead of 3)

app/main.py
```diff
@@ -1 +1,14 @@
...
+        os.rename(command[1], command[2])
+
+    patch = "/".join(command[2].split('/')[0:-1])
```
YukihiroSM, line 9: don't split by `/`. Try to use os.path.split.

YukihiroSM (changes requested):
Fix tests and take a look into comments, please.

app/main.py
```diff
@@ -1 +1,14 @@
...
+    command_list = command.split()
+    if len(command_list) < 3 or command_list[0] != "mv":
+        print("command wrong")
+    if "/" not in command_list[2]:
+        os.rename(command_list[1], command_list[2])
+
+    patch, name = os.path.split(command_list[2])
+    os.rename(command[1], name)
+    os.makedirs(patch)
+    os.replace(name, command_list[2])
```
YukihiroSM, line 14: If the command list length will be less then 3 -> we will only print something and continue. After that on the line 11 we will get failed. Add some returns where it's needed.

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1 +1,18 @@
...
+        return
+
+    if "/" not in command_list[2]:
+        os.rename(command_list[1], command_list[2])
+        return
+
+    patch, name = os.path.split(command_list[2])
+    os.rename(command_list[1], name)
+    os.makedirs(patch)
+    os.replace(name, command_list[2])
```
Bohdan-Salamakha, line 18: what are command_list[1] and command_list[2]?
create variables with declarative names

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #329, opened 2023-06-22
Solution

arturiermolenko:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_name, file_name, path_to_new_file = command.split()
```
olena-hul, line 5: Validate command length before splitting

app/main.py
```diff
@@ -1 +1,11 @@
...
+    if command_name == "mv" and os.path.isfile(file_name):
+        new_file = os.path.dirname(path_to_new_file)
+        if new_file:
+            if not os.path.exists(new_file):
+                os.makedirs(new_file)
```
olena-hul, line 10:
```suggestion
            os.makedirs(new_file, exist_ok=True)
```

olena-hul (changes requested):

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if len(command) > 0:
```
olena-hul, line 5: You should validate then command length is equal exactly to 3 parts, because you're performing unpacking with 3 elements

app/main.py
```diff
@@ -1 +1,13 @@
...
+        command_name, file_name, path_to_new_file = command.split()
+        if command_name == "mv" and os.path.isfile(file_name):
+            new_file = os.path.dirname(path_to_new_file)
+            if new_file:
+                if not os.path.exists(new_file):
+                    os.makedirs(new_file)
```
olena-hul, line 11: This is not needed anymore
Move `os.makedirs(new_file, exist_ok=True)` inside `if block`

`exist_ok` parameter automatically checks if the directory exists, so you don't need to perform the check manually

olena-hul (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #335, opened 2023-07-04
Implementad py-move-file

esemesenko:


app/main.py
```diff
@@ -1 +1,22 @@
...
+
+    if command_name == "mv" and os.path.exists(src_file):
+
+        dest_list = dest_file.split("/")
+        path_dir = ""
+
+        for new_dir in dest_list[:-1]:
+            path_dir = os.path.join(path_dir, new_dir)
+            if not os.path.exists(path_dir):
+                os.mkdir(path_dir)
```
Bohdan-Salamakha, line 16: it's better to use
```makedirs(..., exist_ok=True)```

**Review by Bohdan-Salamakha — CHANGES_REQUESTED** [review]

File: app/main.py [hunk 2]
```diff
@@ -1 +1,21 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+
+    command_name, src_file, dest_file = command.split(" ")
```
YukihiroSM, line 6: You will get an exception if length of elements in command is less than 3.

app/main.py
```diff
@@ -1 +1,21 @@
...
+
+    if command_name == "mv" and os.path.exists(src_file):
+
+        dest_list = dest_file.split("/")
```
YukihiroSM, line 10: It's better to use os.path.split. For example, on Windows we have `\` instead of `/`

YukihiroSM (changes requested):
LGTM

app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+
+    command_list = command.split(" ")
```
olena-hul, line 6:
```suggestion
    command_list = command.split()
```

olena-hul (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #338, opened 2023-07-05
'Solution'

ilizh:


Bohdan-Salamakha (changes requested):
fix flake8

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    parts = command.split(" ")
```
EdAlekseiev, line 5: You don't need to pass `" "` to `split` because it is the default value

app/main.py
```diff
@@ -1 +1,15 @@
...
+    source_path = os.path.join(os.getcwd(), parts[1])
+    destination_path = os.path.join(os.getcwd(), parts[2])
```
EdAlekseiev, line 7: Check that command contains 3 values and the first parameter is `mv `

EdAlekseiev (changes requested):

Bohdan-Salamakha (approved):

PR #339, opened 2023-07-05
'Solution'

Kavbian:


app/main.py
```diff
@@ -1 +1,25 @@
...
+        for drc in dir_list:
+            pth += drc
+            if path.exists(pth):
+                pth += "/"
+                continue
+            mkdir(pth)
+            pth += "/"
+
+    with open(file_in, "r") as f_in, open(file_out, "w") as f_out:
+        f_out.write(f_in.read())
```
vsmutok, line 23: bad variable names

vsmutok: use `os.makedirs`

app/main.py
```diff
@@ -1 +1,25 @@
...
+from os import mkdir, remove, path
+
+
+def move_file(command: str) -> None:
+    sep_command = command.split()
+
+    file_in = sep_command[1]
+    file_out = sep_command[2]
+    dir_list = sep_command[2].split("/")
+    dir_list.pop()
```
vsmutok, line 10: check if the command is passed correctly

vsmutok: if an error occurs when trying to split the command into three parts

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,28 @@
...
+
+
+def move_file(command: str) -> None:
+    sep_command = command.split()
+
+    if len(sep_command) != 3:
+        print("Incorrect command!")
+
+    file_in = sep_command[1]
+    file_out = sep_command[2]
```
ArsenPidhoretskyi, line 11: First of all. These are not files.
```suggestion
    cmd, source_path, destination_path = sep_command
```

app/main.py
```diff
@@ -1 +1,28 @@
...
+
+    if "/" in command:
+        path_file_out = ""
+        for drc in dir_list:
+            path_file_out += drc
+            if path.exists(path_file_out):
+                path_file_out += "/"
+                continue
+            makedirs(path_file_out)
+            path_file_out += "/"
```
ArsenPidhoretskyi, line 23: So hard code. Use os.path.split. Also you can do:
makedirs(*list_of_dirs, exist_ok=True)

ArsenPidhoretskyi (changes requested):
https://github.com/Kavbian/py-move-file/blob/develop/checklist.md
Before do task read checklist
Check that files are not the same. And that cmd is "mv"

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from os import makedirs, remove, path
+
+
+def move_file(command: str) -> None:
+    sep_command = command.split()
```
vasylhnatiuk, line 5: sep_command is a bad name

app/main.py
```diff
@@ -1 +1,18 @@
...
+
+    if sep_command[0] != "mv":
+        return
+
+    split_path = path.split(sep_command[2])
```
vasylhnatiuk, line 10: add if condition if sep_command has 3 values, because your code will be failed

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):
do it

vasylhnatiuk (approved):

PR #340, opened 2023-07-06
Solution

prykhodkom:


app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command = command.split()
+    if len(command) != 3 or command[0] != "mv":
+        raise Exception("Wrong command")
+
+    file_path = os.path.dirname(command[2])
```
YukihiroSM, line 9: what about: `command_name, source_path, dest_path = command`?

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command = command.split()
```
YukihiroSM, line 5: It's better not to replace input parameter variables. Better use something different for this.

YukihiroSM (changes requested):
Looks good. But try not to use `command[1], command[2]` Better assign some variables as in comment below. It's more readable in that case :)

Bohdan-Salamakha (changes requested):

EdAlekseiev (approved):
Good job

PR #344, opened 2023-07-09
Solution

pnakongit:


app/main.py
```diff
@@ -1 +1,30 @@
-# write your code here
+import os
+import shutil
+
+
+def move_file(command: str) -> None:
+    commands = command.split(" ")
```
vsmutok, line 6:
```suggestion
    commands = command.split()
```

app/main.py
```diff
@@ -1 +1,30 @@
-# write your code here
+import os
+import shutil
```
vsmutok, line 2: don't use `shutil` implement your solution

app/main.py
```diff
@@ -1 +1,30 @@
...
+        os.rename(old_file, new_file)
+        return None
+
+    new_file_path_list = new_file.split("/")
+
+    directory_path = os.getcwd()
+    for directory in new_file_path_list[:-1]:
+        directory_path = os.path.join(directory_path, directory)
+        if not os.path.exists(directory_path):
+            os.mkdir(directory_path)
```
vsmutok, line 27: use `os.makedirs(exist_ok=True)`

vsmutok (changes requested):

app/main.py
pnakongit: Done

app/main.py
pnakongit: Ok. Chenged soltuion.

app/main.py
pnakongit: Thanks for the elegant solution!

Grayder0152 (approved):

PR #347, opened 2023-07-13
Create 'move_file' function

maxlapchuk:


Grayder0152 (changes requested):

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from os import makedirs, remove
+
+
+def move_file(command: str) -> None:
+    if len(command.split()) == 3 and command.split()[0] == "mv":
+        command, filename, new_file_path = command.split()
+
+        if len(new_file_path.split("/")) > 1:
```
ArsenPidhoretskyi, line 8: Use os.path.split

app/main.py
```diff
@@ -1 +1,20 @@
...
+            makedirs("/".join(new_file_path.split())[:-1], exist_ok=True)
```
ArsenPidhoretskyi, line 9: use os.path.join because in different OS path will be build differently

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (changes requested):

PR #353, opened 2023-07-17
solution

vkleshko:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command = command.split()
+    if len(command) == 3 and command[0] == "mv":
+        cmd, file_old, file_path = command
+        directory, file_name = os.path.split(file_path)
```
ArsenPidhoretskyi, line 8: Also check that file_old and file_path is not the same

ArsenPidhoretskyi (changes requested):

PR #354, opened 2023-07-18
Solution

tanya-burdenko:


app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    cmd, original, file_path = command.split()
```
vasylhnatiuk, line 5: add if condition that command.split() has 3 arguments

vasylhnatiuk (changes requested):

Grayder0152 (approved):

PR #355, opened 2023-07-19
solution

mwellick:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if len(command.split()) != 3:
+        return
+    if command.split()[0] != "mv":
+        return
+    cmd, file_from, file_to = command.split()
```
YukihiroSM, line 9: You are doing split 3 times. What about creating a new variable for splitted command, and working with that variable in all these lines?

app/main.py
```diff
@@ -1 +1,18 @@
...
+    if "/" in file_to:
```
YukihiroSM, line 10: Is this is Windows - we will have `\` in the path. Please, try not to use this. Think about different variations. What about os.path.dirname

YukihiroSM (changes requested):
Overall the solution is good.

app/main.py
mwellick: I have to add "\" in condition to check if it's the path to directory (on windows) But how to implement if? Because this symbol is already reserved by python as a string symbol

mwellick: ok,googled it.
I hope everything is correct

app/main.py
```diff
@@ -1 +1,18 @@
...
+
+
+def move_file(command: str) -> None:
+    splitted_cmd = command.split()
+    if len(splitted_cmd) != 3:
+        return
+    if splitted_cmd[0] != "mv":
+        return
+    cmd, file_from, file_to = splitted_cmd
+    if "/" in file_to or r"\\" in file_to:
```
olena-hul, line 11: You can use os.path.split which will split the path to 2 parts: directories and filename. If the first part is not empty it means that the path contains directories. So you can avoid this manual check of "/" and "\"

olena-hul (changes requested):

YukihiroSM (approved):
Good job!

PR #356, opened 2023-07-19
Solution

numerodiciannove:


app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    cmd, source_path, destination_path = command.split()
```
vsmutok, line 5: Before you must check that command.split() length == 3

app/main.py
```diff
@@ -1 +1,19 @@
...
+    if cmd != "mv":
+        raise ValueError('Invalid command. Only "mv" command is supported.')
+
+    if source_path == destination_path:
+        print("Source file and destination file have the same name. "
+              "No action needed.")
+        return
+
+    if new_file_path and not os.path.exists(new_file_path):
+        os.makedirs(new_file_path)
```
vsmutok, line 17:
```suggestion
        os.makedirs(new_file_path, exist_ok=True)
```

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    cmd, source_path, destination_path = command.split()
+    new_file_path = "/".join(destination_path.split("/")[:-1])
```
vsmutok, line 6: use os.path.join because in different OS path will be build differently

vsmutok: Use os.path.split

vsmutok (changes requested):

app/main.py
vsmutok: also check if `cmd == "mv"`

numerodiciannove: > also check if `cmd == "mv"`

string 8,9:
    if cmd != "mv":
        raise ValueError('Invalid command. Only "mv" command is supported.')

numerodiciannove: > Before you must check that command.split() length == 3

Why? If the length of command.split() is not equal to 3, a ValueError will automatically be raised: "not enough values to unpack (expected 3, got 2 or 1, 0...)" or "ValueError: too many values to unpack", and further execution of the code will be impossible.

ArsenPidhoretskyi: Okay, for you is it an understandable exception? As for me no, we need provide an informative description

ArsenPidhoretskyi (changes requested):

app/main.py
ArsenPidhoretskyi: Validation must be before unpacking

ArsenPidhoretskyi (changes requested):
Changes requested

vasylhnatiuk (approved):

PR #370, opened 2023-08-10
Develop

AndriiIshchenko:


app/main.py
```diff
@@ -1 +1,27 @@
...
+
+def move_file(command: str) -> None:
+
+    if len(command.split()) == 3:
+        action, file_name, path = command.split()
+        if action == "mv":
+            with open(file_name) as file_to_read:
+                if "/" in path:
+                    list_of_directories = path.split("/")[:-1]
+                    file_name_to_move = path.split("/")[-1]
```
vsmutok, line 12: use os.path.split

app/main.py
```diff
@@ -1 +1,27 @@
...
+                    current_path = os.getcwd()
+                    for folder in list_of_directories:
+                        current_path = os.path.join(current_path, folder)
+                        if not os.path.isdir(current_path):
+                            os.mkdir(current_path)
```
vsmutok, line 17: use os.makedirs

vsmutok (changes requested):

app/main.py
AndriiIshchenko: In conditions to this task noted:
Note: You can create directory by os.mkdir("first") from module os. But to create os.mkdir("first/second"), first directory needs to exist. Function should work with different directory depth. Use os.remove() to remove the file.

if use **os.makedirs** we can ommit for loop at all. So I cant use os.path.split in this case I need all folders separetely

Grayder0152: Yes, but `use os.makedirs` can help you with in this case

app/main.py
```diff
@@ -1 +1,27 @@
...
+
+                    with (open(os.path.join(current_path,
+                                            file_name_to_move), "w")
+                          as file_to_write):
+                        file_to_write.write(file_to_read.read())
+
+                else:
+                    with open(path, "w") as file_to_write:
+                        file_to_write.write(file_to_read.read())
```
Grayder0152, line 26: dont repeat yourself

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1 +1,23 @@
...
+        if action == "mv":
+            head, tail = os.path.split(path)
+            with open(file_name) as file_to_read:
+                current_path = os.getcwd()
+                if head:
+                    try:
+                        current_path = os.path.join(current_path, head)
+                        os.makedirs(current_path)
+                    except OSError:
+                        pass
```
ArsenPidhoretskyi, line 17: Use 
os.makedirs(current_path, exist_ok) instead of try-except

app/main.py
```diff
@@ -1 +1,23 @@
...
+
+                with (open(os.path.join(current_path, tail), "w")
```
ArsenPidhoretskyi, line 19:
```suggestion
                with (open(path), "w")
```

ArsenPidhoretskyi, line 19: Also try to use only one with

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+
+    if len(command.split()) == 3:
+        action, file_name, path = command.split()
+        if action == "mv":
```
ArsenPidhoretskyi, line 8:
```suggestion
        if action == "mv" and file_name != path:
```

ArsenPidhoretskyi: Also check that files are not the same

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #371, opened 2023-08-11
solution

neostyle88:


app/main.py
```diff
@@ -1 +1,28 @@
...
+def move_file(command: str) -> None | str:
+    if len(command.split()) != 3:
+        return "Incorrect statement printed"
+
+    cmd, curent_file, destination = command.split()
+
+    if cmd != "mv":
+        return "Command is incorrect"
+
+    dir_list = destination.split("/")
```
YukihiroSM, line 13: Use os.path.split. Because in different systems there might be different path separator

app/main.py
```diff
@@ -1 +1,28 @@
...
+    new_file_name = dir_list[-1]
+    dir_list.remove(new_file_name)
+
+    try:
+        os.makedirs("/".join(dir_list))
```
YukihiroSM, line 18: Use os.path.join

app/main.py
```diff
@@ -1 +1,28 @@
...
+    except FileExistsError:
+        print("Folder is already exists")
```
YukihiroSM, line 20: Use os.makedirs(path, exist_ok=True) to ignore this exceptions

YukihiroSM (changes requested):

EdAlekseiev (approved):
Good job

PR #373, opened 2023-08-12
'Solution'

usKos:


app/main.py
```diff
@@ -1 +1,32 @@
...
+                        continue
+                elif "." in elem:
+                    with (open(file1, "r") as file_in,
+                            open(file2, "w") as file_out):
+                        file_out.write(file_in.read())
+                else:
+                    with (open(file1, "r") as file_in,
+                            open(file1, "w") as file_out):
+                        file_out.write(file_in.read())
+
```
YukihiroSM, line 31: This else does nothing, isn't it? You have 2 different checks before in if and elif, so no other variants are here.

app/main.py
```diff
@@ -1 +1,32 @@
...
+    if len(args) == 3:
+        cmd, file1, file2 = args
+
+        if cmd == "mv" and file1 != file2:
+            dir_path = file2.split("/")
+            path = ""
+
+            for elem in dir_path:
+                if "." not in elem:
+                    path += elem + "/"
```
YukihiroSM, line 16: Try to use os.path.join

app/main.py
```diff
@@ -1 +1,32 @@
...
+
+
+def move_file(command: str) -> None:
+    args = command.split(" ")
+
+    if len(args) == 3:
+        cmd, file1, file2 = args
+
+        if cmd == "mv" and file1 != file2:
+            dir_path = file2.split("/")
```
YukihiroSM, line 11: os.path.split, instead of split

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    args = command.split(" ")
```
EdAlekseiev, line 5: `" "` is default value for `split` method

EdAlekseiev, line 5:
```suggestion
    args = command.split()
```

app/main.py
```diff
@@ -1 +1,31 @@
...
+            path_file2 = ""
+
+            for elem in ls_dir_file2:
+                item.append(elem)
+                path_file2 = os.path.join(*item)
+
+                try:
+                    os.mkdir(path_file2)
+                except WindowsError:
+                    continue
```
EdAlekseiev, line 24: You can use `os.makedirs` with the argument `exist_ok=True` to avoid exception here

EdAlekseiev (changes requested):

YukihiroSM (approved):
Good job! THanks for fixing this!

PR #374, opened 2023-08-12
develop

uzlss:


uzlss:
Also, I would like to say, that there is one Error in pytests. Basically, one test creates "dir" directory and doesn't delete it, if wasn't passed, after it another test creates "dir" and makes an Error, so you need to delete it after.

vasylhnatiuk (approved):

PR #383, opened 2023-08-21
Solution

denys-source:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    keyword, source, target = command.strip().split()
```
Kroll410, line 5: Handle invalid number of values at unpacking

Kroll410 (changes requested):

Kroll410 (approved):

PR #387, opened 2023-08-22
Task Solved

ORENYT:
You can now move file. Tested manually. Used black formatter.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -3,11 +3,7 @@ import os
 
 def move_file(command: str) -> None:
     command_call, file_to_copy, result_file, *redundant = command.split(" ")
-    if redundant:
-        return
-    if command_call != "mv":
-        return
-    if file_to_copy == result_file:
+    if redundant or command_call != "mv" or file_to_copy == result_file:
         return
     if "/" not in result_file:
         os.rename(file_to_copy, result_file)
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,24 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    list_from_command = command.split(" ")
```
Kroll410, line 5: Unpack values

Kroll410 (changes requested):
Simplify your code with creating directories at the beggining if needed

app/main.py
Kroll410: .

Kroll410 (changes requested):

app/main.py
```diff
@@ -1 +1,27 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_call, file_to_copy, result_file, *redundant = command.split(" ")
+    if redundant:
+        return
+    if command_call != "mv":
```
Kroll410, line 8: You can unite next conditions into one

Kroll410 (changes requested):

app/main.py
ORENYT: Unpacked

app/main.py
ORENYT: United

Kroll410 (approved):

PR #389, opened 2023-08-22
solution implemented, please have mercy on me xD

maxkatkalov:


app/main.py
```diff
@@ -1 +1,41 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    """this IF-block just will rename file without moving somewhere"""
+    parts = command.split()
+
+    if len(parts) == 3 and parts[0] == "mv":
```
Kroll410, line 8: You keep that indent across wholefunction, better do the opposite and return fail

app/main.py
```diff
@@ -1 +1,41 @@
...
+        file_to_move, destination = parts[1], parts[2]
```
Kroll410, line 9: why u can't simply unpack?

app/main.py
```diff
@@ -1 +1,41 @@
...
+        destination_parts = destination.split("/")
+        if len(destination_parts) == 1:
+            os.rename(file_to_move, destination)
+            return
+
+        """this block just will open the existing dir and create new file"""
+        *dirs, name_file_2_create = destination_parts
```
Kroll410, line 16: name_file_2?

Kroll410 (changes requested):
Wait a sec :)
You can gradually simplify your code. Handle creation any folders at the beginning and process moving your file. Your solution is overcomplicated

app/main.py
Kroll410: .

Kroll410 (changes requested):

Ihor-MA (approved):

Kroll410 (approved):

PR #390, opened 2023-08-22
Solution py-move-file

katryana:


app/main.py
```diff
@@ -1 +1,19 @@
...
+
+        if directories:
+            os.makedirs(directories, exist_ok=True)
+
+        with (
+            open(moving_file, "r") as file_in,
+            open(path_to_file, "w") as file_out
+        ):
+            file_out.write(file_in.read())
+        os.remove(moving_file)
```
Kroll410, line 19: what if not exist for some reason?

Kroll410 (changes requested):

PR #391, opened 2023-08-23
Develop

ostboiko:


app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> str:
+
+    parts = command.split()
+
+    if len(parts) < 3:
+        return "Invalid command. Please provide source and destination paths."
```
Bohdan-Salamakha, line 9: u don't need to return a string here
just print the information and return nothing
```return```

File: app/main.py [hunk 2]
```diff
@@ -1 +1,22 @@
...
+
+    if destination.endswith("/"):
+        destination = os.path.join(destination, os.path.basename(source_file))
+
+    try:
+        os.makedirs(os.path.dirname(destination), exist_ok=True)
+        os.rename(source_file, destination)
+        return "File moved successfully."
+    except Exception as e:
+        return f"An error occurred: {str(e)}"
```
Bohdan-Salamakha, line 22: u don't need to return a string here
just print the information and return nothing
```return```

**Review by Bohdan-Salamakha — CHANGES_REQUESTED** [review]
1
command "mv file.txt file2.txt" should rename the file
https://github.com/mate-academy/py-move-file#:~:text=mv%20file.txt%20file2.txt
image

File: app/main.py [hunk 3]
```diff
@@ -1 +1,14 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if len(command.split()) == 3:
+        cmd, original, file_path = command.split()
```
EdAlekseiev, line 6: Don't split `command` a few times. Create a separate variable for storing result of `command.split()` and use it where you need

EdAlekseiev, line 6: Check that, cmd is equal to `mv`

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
...
+def move_file(command: str) -> None:
+    command_parts = command.split()
+    if len(command_parts) == 3:
+        cmd, original, file_path = command_parts
+        if cmd == "mv":
+            if "/" in file_path:
+                head, tail = os.path.split(file_path)
+                os.makedirs(head, exist_ok=True)
+                with open(original, "r") as file1,\
+                        open(file_path, "w") as file2:
```
YukihiroSM, line 13: Don't use `\` for line separation. Better use `(` `)`

YukihiroSM (changes requested):

PR #392, opened 2023-08-23
Solution

ochernous:


app/main.py
```diff
@@ -1 +1,35 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_arr = command.split()
+    action = command_arr[0]
```
Kroll410, line 6: use unpacking better

app/main.py
```diff
@@ -1 +1,35 @@
...
+    destination = command_arr[2]
+
+    if source == destination and action == "mv":
+        return
+
+    if "/" not in destination and action == "mv":
+        os.rename(source, destination)
+        return
+
+    if destination.endswith("/"):
```
Kroll410, line 17: Why do we need this?

app/main.py
```diff
@@ -1 +1,35 @@
...
+def move_file(command: str) -> None:
+    command_arr = command.split()
+    action = command_arr[0]
+    source = command_arr[1]
+    destination = command_arr[2]
+
+    if source == destination and action == "mv":
+        return
+
+    if "/" not in destination and action == "mv":
```
Kroll410, line 13: Can you explain why do you need this condition?

Kroll410 (changes requested):

app/main.py
ochernous: If destination ends with "/" - we just have a path without name of new file.
So, name of new file will be name of old file.
Else, we get name of new file from the end of "destination".

Kroll410 (changes requested):
I guess you can simplify your code with creating all needed directories at the beggining (split('/')..?) and processing it after

PR #393, opened 2023-08-24
Solution

ivanzh13:


app/main.py
```diff
@@ -1 +1,26 @@
...
+class InvalidCommandError(Exception):
+    ...
+
+
+def move_file(command: str) -> None:
+    command_list = command.split()
+    if len(command_list) != 3 or command_list[0] != "mv":
+        raise InvalidCommandError("Unknown command.")
+
+    if "/" in command_list[2]:
```
ArsenPidhoretskyi, line 13: It is bad solution to check by /
Instead use os.path.split()

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,25 @@
...
+def move_file(command: str) -> None:
+    command_list = command.split()
+    if len(command_list) != 3 or command_list[0] != "mv":
+        raise InvalidCommandError("Unknown command.")
+
+    path = ""
+    for folder in command_list[2].split("/")[:-1]:
+        path = os.path.join(path, folder)
+        if not os.path.isdir(path):
+            os.mkdir(path)
```
ArsenPidhoretskyi, line 17: First of all use os.path.split
And then use os.makedirs instead of loop

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,22 @@
...
+
+class InvalidCommandError(Exception):
+    ...
+
+
+def move_file(command: str) -> None:
+    command_list = command.split()
+    if len(command_list) != 3 or command_list[0] != "mv":
+        raise InvalidCommandError("Unknown command.")
+
```
vasylhnatiuk, line 12: add `command_name, origin_path, new_path = command_list` and use for more readable code.

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,25 @@
-# write your code here
+import os
+
+
+class InvalidCommandError(Exception):
+    ...
+
+
+def move_file(command: str) -> None:
+    command_name, origin_path, new_path = command.split()
```
vasylhnatiuk, line 9: you must do it after 11 line because if len(command.split()) != 3 code will be broken

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,27 @@
...
+    ...
+
+
+def move_file(command: str) -> None:
+    command_list = command.split()
+
+    if len(command_list) != 3 or command_list[0] != "mv":
+        raise InvalidCommandError("Unknown command.")
+
+    _, origin_path, new_path = command.split()
```
vasylhnatiuk, line 14:
```suggestion
    _, origin_path, new_path = command_list
```

vasylhnatiuk: better to reuse data

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #394, opened 2023-08-24
Solution

VasylTurok:


Polyakiv-Andrey (changes requested):

app/main.py
```diff
@@ -1 +1,38 @@
...
+    action = parts[0]
+    source_file = parts[1]
+    destination = parts[2]
+
+    if source_file == destination and action == "mv":
+        return
+
+    if "/" not in destination and action == "mv":
+        os.rename(source_file, destination)
+        return
```
Dimosphen1, line 25: There is no need to perform the check for the action that equals `"mv"` it was already done in the first if condition.

app/main.py
```diff
@@ -1 +1,38 @@
...
+        )
+
+    action = parts[0]
+    source_file = parts[1]
+    destination = parts[2]
+
+    if source_file == destination and action == "mv":
+        return
+
+    if "/" not in destination and action == "mv":
```
Dimosphen1, line 23: Do not bind script to file systems using "/" as a directory delimiter, find a cross-system way to validate that the destination contains folders.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,38 @@
...
+
+    if destination == "":
+        os.rename(source_file, new_file_name)
+        return
+
+    os.makedirs(destination, exist_ok=True)
+
+    with (
+        open(source_file, "r") as old_file,
+        open(parts[2], "w") as new_file
```
Dimosphen1, line 33: You have `parts[2]` written to a variable, use a variable instead here.

app/main.py
```diff
@@ -1 +1,38 @@
...
+
+def move_file(command: str) -> None:
+    parts = command.split()
+
+    if len(parts) < 3 or parts[0] != "mv":
+        raise InvalidCommandFormat(
+            "Please provide a command like 'mv source_file destination_path'"
+        )
+
+    action = parts[0]
```
Dimosphen1, line 16: There is no need for the `action` variable, remove it from here and if condition on line 20, the content of `parts[0]` was already checked in the first if condition.

Dimosphen1 (changes requested):
Several minor changes were requested.

pavlejviki (approved):
Well done

PR #395, opened 2023-08-24
Solution

Kurta97:


app/main.py
```diff
@@ -1 +1,25 @@
...
+import os
+
+
+def move_file(command: str) -> None:
+    command_split = command.split()
+
+    if len(command_split) == 3:
+        command, source_path, new_path = command_split
+
+        if command == "mv":
```
Kroll410, line 10: if not what do we do?

Kroll410 (changes requested):
Check condition

app/main.py
Kroll410: .

Kroll410 (changes requested):

ORENYT (approved):

app/main.py
```diff
@@ -1 +1,27 @@
...
+        if path:
+            os.makedirs(path, exist_ok=True)
+
+        with (
+            open(source_path, "r") as source,
+            open(new_path, "w") as new_path
+        ):
+            new_path.write(source.read())
+
+        os.remove(source_path)
```
Kroll410, line 24: What if file does not exist? deleted from different source

Kroll410 (changes requested):

app/main.py
Kroll410: .

Kroll410 (changes requested):

IvanRamyk (approved):

PR #401, opened 2023-09-12
in progress

d-beschastnyi:


app/main.py
```diff
@@ -1 +1,22 @@
...
+
+
+def move_file(command: str) -> None:
+    parts = command.split()
+
+    if len(parts) == 3 and parts[0] == "mv":
+        old_file, full_path_to_new_file = parts[1:]
+
+        if "/" not in full_path_to_new_file:
+            os.rename(old_file, full_path_to_new_file)
```
ArsenPidhoretskyi, line 11: Use os.path.split

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,22 @@
...
+    parts = command.split()
+
+    if len(parts) == 3 and parts[0] == "mv":
+        old_file, full_path_to_new_file = parts[1:]
+        dirs, new_file = os.path.split(full_path_to_new_file)
+
+        if dirs:
+            os.makedirs(dirs, exist_ok=True)
+        else:
+            os.rename(old_file, new_file)
```
ArsenPidhoretskyi, line 14: else is not correct because we need create dirs, and then also mv file

ArsenPidhoretskyi (changes requested):

vsmutok (approved):
Good Job 👍

PR #402, opened 2023-09-13
Solution

olgierrd:


tests/test_main.py
```diff
@@ -57,6 +57,6 @@ def test_should_create_multiple_directories_when_they_exist(create_file: callabl
     with open("first_dir/second_dir/third_dir/file2.txt", "r") as file_with_content:
         assert file_with_content.read() == "This is some\n content for\n the file."
 
-    assert os.path.exists("file.txt") is False
+    assert os.path.exists("../file.txt") is False
```
pavlejviki, line 61: you should not change the tests

pavlejviki (changes requested):

tests/test_main.py
olgierrd: oops, accidentally, sorry

app/main.py
```diff
@@ -1 +1,23 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    files = command[3:].split(" ")
+    source_name = files[0]
+    dest_path = files[1]
+    if "/" in dest_path:
```
pavlejviki, line 8: Avoid using a hardcoded "/" as a directory separator in the script. instead, use a platform-independent method to ensure the destination includes folders.

app/main.py
```diff
@@ -1 +1,23 @@
...
+            open(source_name, "rb") as src_file,
+            open(f"{dest_dir}/{dest_file}", "wb") as dest_file
+        ):
+            dest_file.write(src_file.read())
+        os.remove(source_name)
+    else:
+        os.rename(source_name, dest_path)
+
+
+# move_file("mv file.txt first/file.txt")
```
pavlejviki, line 23: remove comment

tests/test_main.py
```diff
@@ -42,7 +42,7 @@ def test_should_create_multiple_directories(create_file: callable) -> None:
     move_file(f"mv file.txt first_dir/second_dir/file2.txt")
 
     assert os.path.exists("first_dir/second_dir/file2.txt") is True
-    assert os.path.exists("file.txt") is False
+    assert os.path.exists("../file.txt") is False
 
     with open("first_dir/second_dir/file2.txt", "r") as file_with_content:
         assert file_with_content.read() == "This is some\n content for\n the file."
```
pavlejviki, line 48: please roll back all the changes in the test

pavlejviki (changes requested):

README.md
```diff
@@ -14,6 +14,7 @@ Examples:
 - `mv file.txt first_dir/second_dir/file2.txt` create directory `first_dir`
 inside current directory, then create directory `first_dir/second_dir`,
 create file2.txt inside and remove the source file.
+
 ```python
 print(open("file.txt").read())
```
viktoria-rybenchuk, line 19: Please roll back all the changes in the README.md file. You only need to make changes in the main.py

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    files = command[3:].split(" ")
+    source_name = files[0]
+    dest_path = files[1]
```
viktoria-rybenchuk, line 7: You can use unpacking to simplify the code

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    files = command[3:].split(" ")
```
viktoria-rybenchuk, line 5: Make sure you check that the mv input command is correct

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None | str:
+    if command.startswith("mv "):
+        pass
```
viktoria-rybenchuk, line 6: what if the command is 'mvp'

viktoria-rybenchuk: The pass statement is used as a placeholder when you need something syntactically but don't want it to do anything

app/main.py
```diff
@@ -1 +1,22 @@
...
+    else:
+        return "write correct command"
+    source_name, dest_path = command[3:].split(" ")
```
viktoria-rybenchuk, line 9: You should first check if, after splitting, you have obtained 3 values, and only then can you unpack them like this.

viktoria-rybenchuk (changes requested):

app/main.py
olgierrd: > what if the command is 'mvp'

but I check if it starts with "mv " not "mv"

app/main.py
viktoria-rybenchuk: You should check  `if command == 'mv' ` condition is much clearer and more concise. It explicitly checks if the command variable is exactly equal to "mv,"

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (changes requested):
make changes related to the comments

viktoria-rybenchuk (approved):
Good job!

PR #403, opened 2023-09-14
'Solution 14.09.23'

TymchenkoV:


app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+import os.path
+
+
+def move_file(command: str) -> None:
+    cmd, file, new_file = command.split()
+
+    if len(command.split()) != 3 or cmd != "mv":
```
ArsenPidhoretskyi, line 7: It this case we already have an error:
Check this code:
```
command = "cp"
cmd, file, new_file = command.split()
if len(command.split()) != 3 or cmd != "mv":
     raise ValueError("Error#1")
```
You meed to validate len(command.split()) != 3 before upacking:
```
command = "cp"
parts = command.split()
if len(parts) != 3:
    raise ValueError("Description")

cmd, file, new_file = parts
if cmd != "mv" # also add check that files are not the same:
     raise ValueError("Description")
```

app/main.py
```diff
@@ -1 +1,19 @@
...
+        raise SyntaxError("Incorrect command")
+
+    destination, file_name = os.path.split(new_file)
+
+    if "/" in new_file:
```
ArsenPidhoretskyi, line 12:
```suggestion
    if destination:
```

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #407, opened 2023-09-20
solution

antonnech2309:


app/main.py
```diff
@@ -1 +1,22 @@
...
+def move_file(command: str) -> None:
+    file_names = command.split(" ")
+
+    if file_names[0] != "mv" or len(file_names) != 3:
+        return
+    with open(file_names[1], "r") as file:
+        text = file.read()
+    os.remove(file_names[1])
+
+    file_path = file_names[2].split("/")
```
pavlejviki, line 13: see checklist and refactor your code accordingly

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #409, opened 2023-09-22
Solution

Manynik:


app/main.py
```diff
@@ -1 +1,18 @@
...
+import os
+
+
+def move_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3 or parts[0] != "mv":
+        return
+
+    source = parts[1]
+    target = parts[2]
```
viktoria-rybenchuk, line 10: You can use tuple unpacking to assign values from the `parts`

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3 or parts[0] != "mv":
+        return
```
viktoria-rybenchuk, line 7: You can simplify the task with one `return`

viktoria-rybenchuk (changes requested):
Follow comments and read checklist

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    parts = command.split()
+    source, target = parts[1], parts[2]
```
viktoria-rybenchuk: First, check if the length of `parts` is 3 and you can unpack command as well

app/main.py
```diff
@@ -1 +1,14 @@
...
+    if len(parts) != 3 or parts[0] != "mv" or source == target:
+        return
+
+    if os.path.dirname(target) != "":
```
viktoria-rybenchuk, line 10: You can simply check it with `if target`.

app/main.py
viktoria-rybenchuk:
```suggestion
   command,  source, target = parts
```

viktoria-rybenchuk (changes requested):

app/main.py
Manynik: i don't understand

app/main.py
Manynik: how i can do it with one "return"?

app/main.py
viktoria-rybenchuk: first you need  make this condition
Screenshot from 2023-09-22 14-41-55

viktoria-rybenchuk (changes requested):

PR #410, opened 2023-09-22
Solution

Manynik:


app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    parts = command.split()
+    command, source, target = parts
+    if len(parts) != 3 or parts[0] != "mv" or source == target:
```
olena-hul, line 7: check length of parts before unpacking

olena-hul (changes requested):

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3:
+        return
+    command, source, target = parts
+    if parts[0] != "mv" or source == target:
```
Dimosphen1, line 9: Replace `parts[0]` with the `command` variable, you already performed unpacking. Besides that, check all three conditions in one if condition, there in no need to split them into two if conditions. You could use `any` for that.

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #412, opened 2023-09-27
'Solution_move_file'

nsdNite:


app/main.py
```diff
@@ -1 +1,30 @@
...
+        return func(command)
+
+    return wrapper
+
+
+@check_command
+def move_file(command: str) -> None:
+    source_name, target_name = command.split()[1:3]
+    if "/" not in target_name:
+        os.rename(source_name, target_name)
```
viktoria-rybenchuk, line 22: Instead of relying on a hardcoded '/' as a directory separator in the script, consider using a platform-agnostic approach to ensure that the destination path includes the necessary folders

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,33 @@
...
+            raise CommandError("Please enter valid command!")
+        return func(command)
+
+    return wrapper
+
+
+@check_command
+def move_file(command: str) -> None:
+    _, source_name, target_name = command.split()
+    destination_path = os.path.join(target_name)
```
viktoria-rybenchuk, line 23: why do you need this? you can make it without the need to assign it to destination_path separately

app/main.py
```diff
@@ -1 +1,33 @@
...
+    if not os.path.dirname(destination_path):
+        os.rename(source_name, destination_path)
+    else:
+        path_to_file = os.path.split(target_name)[0]
```
viktoria-rybenchuk, line 27: you have already used `os.path.dirname(target_name)` to extract the directory part of target_name.

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,32 @@
...
+
+
+class CommandError(Exception):
+    """Checks command input for right format."""
+
+
+def check_command(func: Callable) -> Callable:
+    @wraps(func)
+    def wrapper(command: str) -> Any:
+        if command.split()[0] != "mv" or len(command.split()) != 3:
```
Dimosphen1, line 13: You could perform the `split` once and do not use `split` twice in one if condition.

app/main.py
```diff
@@ -1 +1,32 @@
...
+    return wrapper
+
+
+@check_command
+def move_file(command: str) -> None:
+    _, source_name, target_name = command.split()
+    if not os.path.dirname(target_name):
+        os.rename(source_name, target_name)
+    else:
+        path_to_file = os.path.split(target_name)[0]
```
Dimosphen1, line 26: You could perform unpacking of the split like 
```suggestion
        path_to_file, _ = os.path.split(target_name)
```
Without using indexes.

Dimosphen1 (approved):
LGTM, several comments were left for consideration.

app/main.py
nsdNite: should i use additional variable for this case?
e.g.:
 words = command.split()
        if words[0] != "mv" or len(words) != 3:

PR #414, opened 2023-09-28
'Solution'

o-sapohova:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command = command.split()
+    if len(command) == 3:
```
olena-hul, line 6: validate command name here too

app/main.py
```diff
@@ -1 +1,16 @@
...
+        start_command, origin_file, new_file = command
+        if start_command == "mv" and "/" not in new_file:
```
olena-hul, line 8: it is bad practice to check for "/" to determine whether there is directory in path or not
Use os.path.split

olena-hul (changes requested):

Bohdan-Salamakha (approved):

PR #418, opened 2023-10-04
Implemented move_file function

sberdianskyi:


app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if len(command.split()) == 3:
+        cmd, file_old, path = command.split()
```
ArsenPidhoretskyi, line 6: Also check that files not the same and cmd is "mv"

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
```
vasylhnatiuk, line 4:
```suggestion
def move_file(command: str) -> None:
    splitted_command = command.split()
```

app/main.py
```diff
@@ -1 +1,14 @@
...
+    if len(command.split()) == 3 and command.split()[0] == "mv":
```
vasylhnatiuk, line 5:
```suggestion
    if len(splitted_command) == 3 and splitted_command[0] == "mv":
```

vasylhnatiuk (changes requested):

ArsenPidhoretskyi (approved):
Good job! You can remove prints

PR #421, opened 2023-10-14
Develop

TROTPOL:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if not command:
+        return
+    command_list = command.split()
+    if command_list[0] == "mv" and len(command_list) == 3:
+        if "/" in command_list[2]:
```
vsmutok, line 9: `mv, file_name, new_file_name = command_list`
use unpacking to avoid code duplication

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if not command:
+        return
+    command_list = command.split()
+    mv, file_name, new_file_name = command_list
+    if mv == "mv" and len(command_list) == 3:
```
vsmutok, line 9: first check that you can unpack, that is, whether the command_list contains three elements, and only then unpack

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,19 @@
...
+def move_file(command: str) -> None:
+    if not command:
+        return
+    command_list = command.split()
+
+    if len(command_list) == 3:
+        mv, file_name, new_file_name = command_list
+        if mv == "mv":
+            if "/" in new_file_name:
+                path_file = new_file_name.split("/")
```
viktoria-rybenchuk, line 13: your app will be cross-platform and will use either / or \ depending on OS. Read checklist. Fix it related to this case

app/main.py
```diff
@@ -1 +1,19 @@
...
+                os.makedirs("/".join(path_file[:-1]), exist_ok=True)
```
viktoria-rybenchuk, line 14: Read checklist, fix here as well

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Well done!

PR #422, opened 2023-10-14
Solution

YuliaHladyshkevych:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    mv_command = command.split()
+    if len(mv_command) == 3 or mv_command[0] == "mv":
```
prochigor, line 6: use  !=, avoid code nesting

prochigor (changes requested):

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    mv_command = command.split()
```

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🍬

PR #425, opened 2023-10-16
Solution

Naz-iv:


D1st3f (approved):

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+from os import path, makedirs, replace
+
+
+def move_file(command: str) -> None:
+    mv_command = command.split()
```

app/main.py
```diff
@@ -1 +1,12 @@
...
+    if mv_command[0] != "mv" or len(mv_command) != 3:
+        return None
```
LiudmylaKulinchenko, line 7: U can use opposite condition here

app/main.py
```diff
@@ -1 +1,12 @@
...
+
+    path_file = path.dirname(mv_command[2])
+    if path_file:
+        makedirs(path_file, exist_ok=True)
+    replace(mv_command[1], mv_command[2])
```
LiudmylaKulinchenko, line 12: `Use os.remove() to remove the file.`

LiudmylaKulinchenko (changes requested):

app/main.py
Naz-iv: done

app/main.py
Naz-iv: fixed

app/main.py
Naz-iv: changed to remove() method

LiudmylaKulinchenko (approved):
Well done! 🌷

PR #426, opened 2023-10-16
solution

EvgenMilokumov:


app/main.py
```diff
@@ -1 +1,26 @@
...
+import shutil
+
+
+def move_file(command: str) -> None:
+    splitted_command = command.split()
+
+    if len(splitted_command) < 3:
+        print(f"Incorrect command {command}")
+        return
+
```
vsmutok, line 11: Make sure you check that the mv input command is correct.

app/main.py
```diff
@@ -1 +1,26 @@
...
+    init_file = splitted_command[1]
+    new_file_directory_with_name = splitted_command[2]
```
vsmutok, line 13: use unpacking

app/main.py
```diff
@@ -1 +1,26 @@
...
+    new_file_directory = os.path.dirname(new_file_directory_with_name)
+
+    if not os.path.exists(init_file):
+        print(f"File: '{init_file}' isn't exist.")
+        return
+
+    try:
+        os.makedirs(new_file_directory)
+        shutil.move(init_file, new_file_directory_with_name)
```
vsmutok, line 22: Don't  use `shutil`, create  your  custom  logic

vsmutok (changes requested):

app/main.py
EvgenMilokumov: done

app/main.py
EvgenMilokumov: done

app/main.py
```diff
@@ -1 +1,23 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+
+    if "mv" not in command or len(command.split()) != 3:
```
ArsenPidhoretskyi, line 6: This is not true because when command will equal to hello mv.txt second.txt it will have incorrect behavior

ArsenPidhoretskyi (changes requested):

app/main.py
EvgenMilokumov: done, thnx

Yurnerroo (approved):

PR #430, opened 2023-10-16
solutio

kostomeister:


vshvanska (approved):

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_list = command.split(" ")
```
LiudmylaKulinchenko, line 5: no need to use space inside split method, it uses space by default

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🌷

PR #431, opened 2023-10-17
Solution

wataschi:


app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    get_command = command.split(" ")
+    if get_command[0] == "mv" and len(get_command) == 3:
+        old_file_name = get_command[1]
+        new_file_name = get_command[2]
```
Dimosphen1, line 8: You could unpack the `get_command` variable on the next line after obtaining it, this approach would improve your code readability by eliminating the need to obtain values via indexes on these lines.

app/main.py
```diff
@@ -1 +1,22 @@
...
+        if "/" in new_file_name:
```
Dimosphen1, line 9: Do not track specific directory delimiters in your code that are specific to the platform. Use built-in methods in `os` to make your code platform-agnostic.

app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    get_command = command.split(" ")
```
Dimosphen1, line 5: There is no need to prove a space as a delimiter, it is a value by default. Remove it.

app/main.py
```diff
@@ -1 +1,22 @@
...
+    if get_command[0] == "mv" and len(get_command) == 3:
+        command, old_file_name, new_file_name = get_command
```
Dimosphen1, line 7: You already performed a command check, there is no need to allocate a variable for that value. Omit it using the following syntax:
```suggestion
        _, old_file_name, new_file_name = get_command
```

app/main.py
```diff
@@ -1 +1,22 @@
...
+        directory = os.path.dirname(new_file_name)
+        if directory:
+            if not os.path.exists(os.path.dirname(new_file_name)):
+                os.makedirs(os.path.dirname(new_file_name), exist_ok=True)
```
Dimosphen1, line 11: You already have the result of calculation in the `directory` variable, replace the following `os.path.dirname(new_file_name)` calls with `directory` variable.

Dimosphen1 (changes requested):

pavlejviki (approved):
Well done!

PR #437, opened 2023-10-20
solution

Stainback:


app/main.py
```diff
@@ -1 +1,46 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    source, destination = read_command(command)
+
+    with open(source, "r") as target_file, \
+         open(destination, "w") as destination_file:
```
YukihiroSM, line 8: Don't use `\` for line separation, Better to use `(` `)`

app/main.py
```diff
@@ -1 +1,46 @@
...
+
+def read_command(command: str) -> tuple[str, str]:
+    command_items = command.split()
+
+    try:
+        verify_command(command_items)
+    except (ValueError, FileNotFoundError) as err:
+        print(err)
+
+    return command_items[1], create_path(command_items[-1].split("/"))
```
YukihiroSM, line 22: use `os.path.split()` instead of splitting by `/`

app/main.py
```diff
@@ -1 +1,46 @@
...
+    os.remove(source)
+
+
+def read_command(command: str) -> tuple[str, str]:
+    command_items = command.split()
+
+    try:
+        verify_command(command_items)
+    except (ValueError, FileNotFoundError) as err:
+        print(err)
```
YukihiroSM, line 20: It whould be better to finish execution in case of any error

YukihiroSM (changes requested):

YukihiroSM (approved):
Looks good to me!

PR #439, opened 2023-10-23
Solution

alecilchenko:


tests/test_main.py
```diff
@@ -27,7 +27,7 @@ def test_file_renamed(create_file: callable) -> None:
 
 
 def test_should_work_when_directory_exists(create_file: callable) -> None:
-    os.makedirs("dir")
+    os.makedirs("dir", exist_ok=True)
```
viktoria-rybenchuk, line 30: rollback changes in tests

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+import os
+import shutil
+
+
+def move_file(command: str) -> None:
+    tag, source, destination = command.split()
```
viktoria-rybenchuk, line 6: First, you need to check that the length of the command is 3

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,19 @@
...
+import os
+import shutil
+
+
+def move_file(command: str) -> None:
+    command_split = command.split()
+    if len(command_split) != 3:
+        return
+    tag, source, destination = command_split
+    if len(destination.split("/")) == 1:
```
pavlejviki, line 10: The code uses a hardcoded forward slash as the path separator. This is common on UNIX and UNIX-like systems, but Windows traditionally uses a backslash . A more cross-platform approach would be to use os.path.join() and os.path.split()

pavlejviki (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #445, opened 2023-10-26
Solution

DanSheremeta:
Solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,26 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    mv, old_filename, new_filename = command.split()
+
+    if mv != "mv":
+        return
+
+    splited_new_filename = new_filename.split("/")
+    if len(splited_new_filename) == 1:
+        os.rename(old_filename, new_filename)
+        return
+
+    dirs = splited_new_filename[:-1]
+    with open(old_filename, "r") as old_file:
+        old_file_content = old_file.read()
+        previous = ""
+        for i, directory in enumerate(dirs):
+            if i != 0:
+                directory = previous + "/" + directory
+            os.makedirs(directory, exist_ok=True)
+            previous = directory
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,26 @@
...
+import os
+
+
+def move_file(command: str) -> None:
+    mv, old_filename, new_filename = command.split()
+
+    if mv != "mv":
+        return
+
+    splited_new_filename = new_filename.split("/")
```
YukihiroSM, line 10: use os.path.split instead of split

app/main.py
```diff
@@ -1 +1,26 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    mv, old_filename, new_filename = command.split()
```
YukihiroSM, line 5: check that command length is 3

app/main.py
```diff
@@ -1 +1,26 @@
...
+        os.rename(old_filename, new_filename)
+        return
+
+    dirs = splited_new_filename[:-1]
+    with open(old_filename, "r") as old_file:
+        old_file_content = old_file.read()
+        previous = ""
+        for i, directory in enumerate(dirs):
+            if i != 0:
+                directory = previous + "/" + directory
```
YukihiroSM, line 21: use os.path.join

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1 +1,29 @@
...
+    path, new_filename = os.path.split(new_filepath)
+
+    if mv != "mv":
+        return
+
+    if path == "":
+        os.rename(old_filename, new_filename)
+        return
+
+    dirs = path.split("/")
```
EdAlekseiev, line 18: Use `os.path.split`

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #448, opened 2023-11-01
Solution

IvanMozhar:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_parts = command.split()
+
+    source = command_parts[1]
+    destination = command_parts[2]
```
Dimosphen1, line 8: You should additionally check that the command itself is `mv`, only in that case logic makes sense.

Dimosphen1, line 8: You could use unpacking to dynamically assign values to variables and avoid using indexes.

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #449, opened 2023-11-01
Vyrvit' meni ochi

hbilous:


app/main.py
```diff
@@ -1 +1,25 @@
...
+        elif "/" in path_file:
+            full_path_file = path_file
+            path_file = path_file.split("/")
+            new_file_name = path_file[-1]
+            path_file.pop(-1)
+            path_file = "/".join(path_file)
+            if not os.path.exists(path_file):
+                os.makedirs(path_file)
+            if os.path.exists(new_file_name):
+                os.remove(new_file_name)
```
NikOneZ1, line 18: Redundant, because in the write mode you will anyway rewrite data from this file

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    mv, source_file, path_file = command.split()
+    if len(command.split()) == 3 and mv == "mv":
```
YukihiroSM, line 6: If len(command.split) will be less, than 3 - it will fail in the previous line. Please, move the if expression to the beginning

app/main.py
```diff
@@ -1 +1,20 @@
...
+        if "/" not in path_file:
+            os.rename(source_file, path_file)
+        elif "/" in path_file:
```
YukihiroSM, line 9: If I'm using Windows - I'm using `\` not `/`

YukihiroSM (changes requested):

ArsenPidhoretskyi (changes requested):

EdAlekseiev (approved):
Good job

PR #451, opened 2023-11-02
Solution

okien1:


app/main.py
```diff
@@ -1 +1,26 @@
...
+import os
+
+
+def move_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) < 3:
+        raise ValueError("Invalid command format.")
+
+    source_file = parts[1]
+    destination = parts[2]
```
Dimosphen1, line 10: Perform unpacking and check that the command name is `mv` and the source filename does not equal the destination filename before moving forward. Don't use indexes to retrieve values here.

app/main.py
```diff
@@ -1 +1,26 @@
...
+    else:
+        destination_path = destination
+
+    # Create directories if they don't exist
+    destination_directory = os.path.dirname(destination_path)
+    if destination_directory:
+        os.makedirs(destination_directory, exist_ok=True)
+
+    # Attempt to move the file
+    os.rename(source_file, destination_path)
```
Dimosphen1, line 26: Remove comments from your pull request, the code is self-descriptive enough.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    comand, source_file, destination = command.split()
```
viktoria-rybenchuk, line 5: you should check length of command before unpacking in order to avoid unexpected error

app/main.py
```diff
@@ -1 +1,19 @@
...
+
+    if comand == "mv" and source_file != destination:
+        if destination.endswith("/"):
```
viktoria-rybenchuk, line 8: you should  to perform path-related operations in a platform-independent way, you can use the os.path module. In this case, you can use os.path.sep to obtain the platform-specific path separator and then check if the destination string ends with that separator.

viktoria-rybenchuk (changes requested):

pavlejviki (approved):
GJ!

PR #453, opened 2023-11-04
Written a function "move_file" that will move a file from one location …

AndriiHomeniuk1:
…to another.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,32 @@
-# write your code here
+import os
+from shutil import move
+
+
+def move_file(command: str) -> None:
+    if len(command.split()) != 3:
+        print(
+            "Error: The command must contain "
+            "three parts separated by spaces."
+        )
+        return
+
+    mv, file_name, move_path = command.split()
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,32 @@
...
+        directory_path = move_path
+        new_file_name = os.path.join(directory_path, file_name)
+    else:
+        directory_path = os.path.dirname(move_path)
+        new_file_name = move_path
+
+    if directory_path and not os.path.exists(directory_path):
+        os.makedirs(directory_path, exist_ok=True)
+
+    move(file_name, new_file_name)
```
ArsenPidhoretskyi, line 32: Please, do not use shutil

ArsenPidhoretskyi (changes requested):

app/main.py
AndriiHomeniuk1: Yes)
I did it another way.

Yurnerroo (approved):

PR #455, opened 2023-11-05
Implemented py-move-file

Aeros72:


app/main.py
```diff
@@ -1 +1,33 @@
...
+        content = old_file.read()
+        previous = ""
+        for i, directory in enumerate(dirs):
+            directory = os.path.join(previous, directory)
+            os.makedirs(directory, exist_ok=True)
+            previous = directory
+
+        with open(destination_path, "w") as new_file:
+            new_file.write(content)
+    os.remove(filename)
```
Dimosphen1, line 33: Omit using context managers to solve this task. You could implement the logic using only `os` methods.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,29 @@
...
+        return
+
+    mv, filename, destination_path = parts
+    path, new_filename = os.path.split(destination_path)
+
+    if mv != "mv":
+        return
+
+    if not os.path.exists(filename):
+        return
```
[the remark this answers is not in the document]
viktoria-rybenchuk: it is considered better because it enhances code readability, reduces nesting, and is more concise while still expressing the same conditions for the file move to occur

app/main.py
```diff
@@ -1 +1,29 @@
...
+
+    if destination_path.endswith("/"):
```
viktoria-rybenchuk, line 19: omit checking explicitly for "/". What if you have different OS?

viktoria-rybenchuk (changes requested):

app/main.py
Aeros72: sorry, but i can`t understand how to do it, can you give me a hint?

app/main.py
```diff
@@ -1 +1,24 @@
...
+    parts = command.split()
+
+    if len(parts) != 3:
+        return
+
+    mv, filename, destination_path = parts
+    new_file_name = ""
+
+    if mv == "mv" and os.path.exists(filename):
+        if destination_path.endswith("/"):
```
Dimosphen1, line 14: You could use `os.path.isdir` to check wether the destination is a directory of not. Omit using `/` to perform this check.

Dimosphen1 (changes requested):

app/main.py
Aeros72: Thanks

pavlejviki (approved):
Well done!

PR #457, opened 2023-11-06
solution

Andrey1104:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    mv, file, directory = command.split()
```
viktoria-rybenchuk, line 5: you should check length of command.split before unpacking in order to avoid unexpected error

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #461, opened 2023-11-09
Solution_1

Taran21V:


app/main.py
```diff
@@ -1 +1,23 @@
...
+            directory = new_file_name
+            update_file_name = os.path.join(new_file_name, file_name)
+        else:
+            directory = os.path.dirname(new_file_name)
+            update_file_name = new_file_name
+
+        if not os.path.exists(new_file_name) and directory:
+            os.makedirs(directory, exist_ok=True)
+
+    os.rename(file_name, update_file_name)
```
NikOneZ1, line 23: Why do you rename file even if the command not equal to mv? It would be better to move this code to the "if os.path.exists(file_name) and com == "mv":" statement

app/main.py
```diff
@@ -1 +1,23 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+
+    sep_command = command.split()
+    if len(sep_command) != 3:
+        raise ValueError
```
NikOneZ1, line 8: It is a good practice to add error message when you raise the error

NikOneZ1 (changes requested):

EdAlekseiev (approved):
Good job

PR #462, opened 2023-11-11
Solution

CrazyHilly:


app/main.py
```diff
@@ -1 +1,109 @@
...
+
+    # Check if input command format is correct
+    if not is_input_data_correct(command, correct_slash):
+        return
+
+    # Get correct paths for source and destination files
+    origin_file_path, dest_file_path = command.split()[1:]
+    if dest_file_path.endswith(correct_slash):
+        origin_file_name = origin_file_path.split(correct_slash)[-1]
+        dest_file_path += origin_file_name
```
NikOneZ1, line 18: You can get rid of the functionality with the slashes and use os.path.split()

NikOneZ1 (changes requested):

app/main.py
```diff
@@ -1 +1,110 @@
...
+        return
+
+    # Get correct paths for source and destination files
+    origin_input_path, dest_input_path = command.split()[1:]
+    dest_short_path, dest_file_name = os.path.split(dest_input_path)
+    origin_short_path, origin_file_name = os.path.split(origin_input_path)
+    if not dest_file_name:
+        dest_full_path = os.path.join(dest_short_path, origin_file_name)
+    else:
+        dest_full_path = dest_input_path
```
ArsenPidhoretskyi, line 16:
```suggestion
        dest_input_path = os.path.join(dest_short_path, origin_file_name)
```
dest_full_path is redundant variable name

app/main.py
```diff
@@ -1 +1,110 @@
...
+    except FileNotFoundError:
+        print(f"No source file with such name: {origin_input_path}")
+        return
+    except PermissionError:
+        print(f'"{origin_file_name}" is a folder')
+        return
+
+    # Create folder tree given in the destination path
+    try:
+        create_folder_tree(dest_short_path)
```
ArsenPidhoretskyi, line 37:
```suggestion
        if dest_short_path:
              os.makedirs(dest_short_path, exist_ok=True)
```
Simplify code please

app/main.py
```diff
@@ -1 +1,110 @@
...
+        print(f'"{dest_file_name}" is a folder')
+    except OSError:
+        print(r'File name cannot contain the following characters: \/?*:"<>|')
+        return
+    else:
+        origin_file.close()
+        os.remove(origin_input_path)
+
+
+def create_folder_tree(folder_path: str) -> None:
```
ArsenPidhoretskyi, line 63: This is redundant

app/main.py
```diff
@@ -1 +1,110 @@
...
+    try:
+        create_folder_tree(dest_short_path)
+    except OSError:
+        print(r'File name cannot contain the following characters: \/?*:"<>|')
+        return
+
+    # Create target file, copy data to it from source file, remove source file
+    dest_full_path = os.path.normpath(dest_full_path)
+    try:
+        with open(dest_full_path, "x") as dest_file:
```
ArsenPidhoretskyi, line 45:
```suggestion
        with open(dest_full_path, "w") as dest_file:
```

ArsenPidhoretskyi (changes requested):

app/main.py
CrazyHilly: thanks!

app/main.py
CrazyHilly: I use "x" to avoid modifying an existing file. Is it expected that if the user types the name of an existing file within the target path, it should be replaced with the source file? If so, it's not quite clear from the task description.

EdAlekseiev: You need to use open(dest_full_path, "w")

app/main.py
EdAlekseiev: Ok

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #463, opened 2023-11-13
Solution

aleksandra-prognimak:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    curr_file, full_path = command.split(" ")[1:]
```
LiudmylaKulinchenko, line 5: You should also check if command is correct (`== "mv"`)

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_, current_file, full_path = command.split()
+    path, new_file = os.path.split(full_path)
+
+    if command_ == "mv":
+        if path:
```
LiudmylaKulinchenko, line 9: U can use `and` to avoid nested conditions

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 🌷

PR #468, opened 2023-11-20
Solution

lyaskov:


app/main.py
```diff
@@ -1 +1,36 @@
-# write your code here
+import os
+
+
+def make_directories(destination: str) -> None:
+    directories = destination.split("/")[:-1]
```
EdAlekseiev, line 5: Use `os.path.split`

app/main.py
```diff
@@ -1 +1,36 @@
...
+    path = ""
+    for directory in directories:
+        path += directory + "/"
+        if not os.path.exists(path):
+            os.mkdir(path)
```
EdAlekseiev, line 10: You can specify this code.
```suggestion
    path = os.path.join(directories)
    os. makedirs(path, exist_ok=True)
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #469, opened 2023-11-22
Solution

Evheniia96:


NikOneZ1 (changes requested):
Check if the command is 'mv' and that command contains 3 parts

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    parts = command.split()
+    if not parts or parts[0] != "mv" or len(parts) != 3:
+        return
+    source_file = parts[1]
+    destination = parts[2]
```
EdAlekseiev, line 9: You can use unpacking
```suggestion
    _, source_file, destination = parts
```

app/main.py
```diff
@@ -1 +1,17 @@
...
+    if parts[1] == parts[2]:
```
EdAlekseiev, line 10:
```suggestion
    if source_file == destination:
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #474, opened 2023-11-25
Solution

Liliyakhu:


app/main.py
```diff
@@ -1 +1,37 @@
...
+            return
+        if not os.path.exists(source):
+            return
+
+        linux = destination.count("/")
+        windows = destination.count("\\")
+        if linux > windows:
+            dir_separator = "/"
+        else:
+            dir_separator = "\\"
```
viktoria-rybenchuk, line 20: use ` os.path.sep` instead of this code

app/main.py
```diff
@@ -1 +1,37 @@
...
+def move_file(command: str) -> None:
+    try:
+        mv, source, destination = command.split()
+    except ValueError:
+        print("Wrong argument")
+    else:
+        if mv != "mv":
+            return
+        if not os.path.exists(source):
+            return
```
viktoria-rybenchuk, line 13: You can combine the conditions using the logical `or` operator

app/main.py
```diff
@@ -1 +1,37 @@
...
+        linux = destination.count("/")
+        windows = destination.count("\\")
+        if linux > windows:
+            dir_separator = "/"
+        else:
+            dir_separator = "\\"
+
+        to_join = ""
+        dirs = destination.split(dir_separator)[:-1]
+        file_name = destination.split(dir_separator)[-1]
```
viktoria-rybenchuk, line 24: use unpacking here

app/main.py
```diff
@@ -1 +1,37 @@
...
+
+        for directory in range(len(dirs)):
```
viktoria-rybenchuk, line 26: why do you need `for loop`.  You can simply use `os.path.join`

app/main.py
```diff
@@ -1 +1,37 @@
...
+            new_directory = os.path.join(to_join, dirs[directory])
+            if not os.path.exists(new_directory):
+                os.mkdir(new_directory)
```
viktoria-rybenchuk, line 29: `exist_ok=True` ensures that the operation doesn't raise an error if the directory already exists

viktoria-rybenchuk (changes requested):

Dimosphen1 (changes requested):

PR #475, opened 2023-11-26
Move-File-Solution

MykolaMazh:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+import shutil
```
ArsenPidhoretskyi, line 1: No need to use shutil

ArsenPidhoretskyi (changes requested):

app/main.py
MykolaMazh: I'd like to find out why
```python
import shutil
shutil.move(file_to_move_path, new_file_path)
```
is worse than?
```python
with (open(file_to_move_path, "r") as f_to_read,
      open(new_file_path, "w") as f_to_write):
    f_to_write.write(f_to_read.read())
os.remove(file_to_move_path)
```
or there is a better way

ArsenPidhoretskyi (approved):
Well done!

PR #476, opened 2023-11-27
Solution

maxvell4:


app/main.py
```diff
@@ -1 +1,34 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+
+    command = command.split()
+    print(command)
```
viktoria-rybenchuk, line 7: remove redundant `print`

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+    if len(command) < 3 or command[0] != "mv":
+        raise ValueError
+
+    source_path = command[1]
+    destination_path = command[2]
```
viktoria-rybenchuk, line 13: You can use unpacking to make the code more concise and readable.

app/main.py
```diff
@@ -1 +1,34 @@
...
+            destination_file.write(source_file.read())
+
+        os.remove(f"{source_file.name}")
+        return
+
+    with (open(f"{source_path}", "r") as source_file,
+          open(destination_file_name, "w") as destination_file):
+
+        destination_file.write(source_file.read())
+        os.remove(f"{source_file.name}")
```
viktoria-rybenchuk, line 34: You can simplify the code further by using `os.replace` to handle the file move operation.

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,28 @@
...
+
+    if separator not in destination_path:
+        os.replace(source_path, destination_path)
+        return
+
+    with (open(f"{source_path}", "r") as source_file,
+          open(destination_file_name, "w") as destination_file):
+
+        destination_file.write(source_file.read())
+        os.remove(f"{source_file.name}")
```
Dimosphen1, line 28: It seems that there is no need to wrap `source_file.name` inside of the `f""` because `name` will return string anyway.

app/main.py
```diff
@@ -1 +1,28 @@
...
+def move_file(command: str) -> None:
+
+    command = command.split()
+
+    if len(command) < 3 or command[0] != "mv":
+        raise ValueError
+
+    command, source_path, destination_path = command
+
+    separator = os.path.sep
```
Dimosphen1, line 13: When the `separator` variable is used only once in the logic, it would be preferable to call `os.path.sep` directly where needed.

Dimosphen1 (approved):
LGTM, several comments were left for consideration.

PR #477, opened 2023-11-27
'Solution'

yul1valko:


app/main.py
```diff
@@ -1 +1,30 @@
...
+
+            with open(directories, "w") as dest_file:
+                dest_file.write(user_input)
+
+        else:
+            new_filename = directories
+            with open(filename, "r") as source_file:
+                user_input = source_file.read()
+
+            with open(new_filename, "w") as dest_file:
```
viktoria-rybenchuk, line 27: don't overuse context manager. You can use `os.path.replace` for moving file

app/main.py
```diff
@@ -1 +1,30 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    user_input = command.split()
+    if len(user_input) == 3:
+        filename, directories = user_input[1], user_input[2]
+
+        if "/" in directories:
```
viktoria-rybenchuk, line 9: what if it is another OS? It's recommended to use `os.path.sep` for cross-platform compatibility.

app/main.py
```diff
@@ -1 +1,30 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    user_input = command.split()
+    if len(user_input) == 3:
+        filename, directories = user_input[1], user_input[2]
```
viktoria-rybenchuk, line 7: where do you check that command is exactly `mv`?

viktoria-rybenchuk (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #478, opened 2023-11-28
So;ution for py_move_file

VladyslavBazhyn:


app/main.py
```diff
@@ -1 +1,24 @@
...
+        if action == "mv":
+            file_out_path = file_out.split("/")[:-1]
+            file_in_content = open(file_in, "r").read()
+            os.remove(file_in)
+
+            if len(file_out_path) > 0:
+
+                for i in range(1, len(file_out_path) + 1):
+                    try:
+                        os.mkdir("/".join(file_out_path[:i]))
```
EdAlekseiev, line 19: Use os.path.join instead of "/".join

EdAlekseiev (changes requested):

PR #480, opened 2023-11-29
Solution_29112023

Anton-Konyk:


app/main.py
```diff
@@ -1 +1,40 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    parts = command.split()
+
+    if (parts[0] == "mv"
+       and parts[1] != parts[2]
+       and "/" not in parts[2]
```
EdAlekseiev, line 9: If you want to check that `parts[2]` contains, use `os.path.dirname`.

app/main.py
```diff
@@ -1 +1,40 @@
...
+                    os.chdir(directory)
+                else:
+                    os.mkdir(directory)
+                    os.chdir(directory)
+
+            with open(new_file, "w") as new_file:
+                new_file.write(file_in.read())
+                file_in.close()
+                os.chdir(current_dir)
+                os.remove(parts[1])
```
EdAlekseiev, line 40: I think, this is redundant code. You can create dir inside first if-block

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    parts = command.split()
+
+    if (parts[0] == "mv"
+       and parts[1] != parts[2]
+       and len(parts) == 3):
```
ArsenPidhoretskyi, line 9: and works as lazy conditions. It works to all true of first false, in this case parts[0] can raise error if parts is empty list
```suggestion
    if (
       len(parts) == 3
       and parts[0] == "mv"
       and parts[1] != parts[2]
   ):
```

app/main.py
```diff
@@ -1 +1,29 @@
...
+
+        current_dir = os.getcwd()
+
+        with open(parts[1], "r") as file_in:
+            dirs = parts[2].split("/")
```
ArsenPidhoretskyi, line 14: Use os.path.split

app/main.py
```diff
@@ -1 +1,29 @@
...
+            new_file = dirs.pop(len(dirs) - 1)
+
+            for directory in dirs:
+                if os.path.isdir(directory):
+                    os.chdir(directory)
+                else:
+                    os.mkdir(directory)
+                    os.chdir(directory)
```
ArsenPidhoretskyi, line 22: Use os.makedirs with exist_ok=True instead of so complicated code

ArsenPidhoretskyi, line 22: also if and else have the same code

app/main.py
```diff
@@ -1 +1,29 @@
...
+
+            with open(new_file, "w") as new_file:
+
+                new_file.write(file_in.read())
+                file_in.close()
+                os.chdir(current_dir)
+                os.remove(parts[1])
```
ArsenPidhoretskyi, line 29: This must be removed after context manager

EdAlekseiev (approved):
Good job

PR #481, opened 2023-11-29
'solution'

boris-devs:


app/main.py
```diff
@@ -1 +1,15 @@
...
+
+    if "/" in destination:
+        os.makedirs(os.path.dirname(destination), exist_ok=True)
+
+        with open(source, "r") as file_in, open(destination, "w") as file_out:
+            content = file_in.read()
+            file_out.write(content)
+        os.remove(source)
+    else:
+        os.rename(source, destination)
```
EdAlekseiev, line 15: This check is incorrect because difference OS has difference path separators.
```suggestion
    dir = os.path.dirname(destination)
    if dir:
        os.makedirs(dir, exist_ok=True)

        with open(source, "r") as file_in, open(destination, "w") as file_out:
            content = file_in.read()
            file_out.write(content)
        os.remove(source)
    else:
        os.rename(source, destination)
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
good job

PR #485, opened 2023-12-05
Solution

eLQeR:


app/main.py
```diff
@@ -1 +1,30 @@
...
+from os import mkdir, remove
+
+
+def move_file(command: str) -> None:
+    if command.startswith("mv "):
+        if "/" in command:
+            file_name = command.split()[1]
+            paths = command.split()[2]
+            command = ""
+            for directory in paths.split("/"):
```
EdAlekseiev, line 10: Use `os.path.split` instead if `.split("/")`

app/main.py
```diff
@@ -1 +1,30 @@
...
+                command += directory
+                if "." not in command:
+                    try:
+                        mkdir(command)
+                    except FileExistsError:
+                        pass
+
+                command += "/"
```
EdAlekseiev, line 18: Why?

app/main.py
```diff
@@ -1 +1,30 @@
...
+                  open(paths, "w") as file_copy):
+                file_copy.write(file_to_copy.read())
+            remove(file_name)
+        else:
+            file_name = command.split()[1]
+            final_name = command.split()[2]
+            with (open(file_name, "r") as file_to_copy,
+                  open(final_name, "w") as file_copy):
+                file_copy.write(file_to_copy.read())
+            remove(file_name)
```
EdAlekseiev, line 30: It looks the same as code above. Can you join them?

app/main.py
```diff
@@ -1 +1,30 @@
...
+            file_name = command.split()[1]
+            paths = command.split()[2]
+            command = ""
+            for directory in paths.split("/"):
+                command += directory
+                if "." not in command:
+                    try:
+                        mkdir(command)
+                    except FileExistsError:
+                        pass
```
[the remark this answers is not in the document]
eLQeR: Так навіть makedirs вилітає ця помилка -_-

NikOneZ1 (changes requested):
Add a check that the command consists of three parts

EdAlekseiev (approved):
Good job

PR #486, opened 2023-12-09
Solution

Ded31:


app/main.py
```diff
@@ -1 +1,26 @@
-# write your code here
+from os import makedirs, remove, rename
+
+
+def move_file(command: str) -> None:
+    com, old_filename, path_and_new_filename = command.split()
+    new_filename = path_and_new_filename.split("/")[-1]
```
EdAlekseiev, line 6: Each OS has difference path separator, so you can't use `/` split. Use `os.path.split` instead of `.split("/")`

app/main.py
```diff
@@ -1 +1,26 @@
...
+    full_path = ""
+    for path in path_and_new_filename.split("/")[:-1]:
+        full_path += (path + "/")
```
EdAlekseiev, line 9: The same comment as above

app/main.py
```diff
@@ -1 +1,26 @@
...
+
+    if com != "mv":
+        return
```
EdAlekseiev, line 12: I think, it must be in top of function

app/main.py
```diff
@@ -1 +1,26 @@
...
+
+    if path_and_new_filename == new_filename:
+        rename(old_filename, new_filename)
+        return
+
+    try:
+        makedirs(full_path)
```
EdAlekseiev, line 19: You can use makedirs(full_path, exist_ok=True) to avoid `FileExistsError`

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #492, opened 2023-12-12
implemented move_file function

380974261314kv:


app/main.py
```diff
@@ -1 +1,29 @@
...
+                    print("Such a directory has already exsists")
+        else:
+            print("Error: Operation should be 'mv'")
+    else:
+        print("Error: Invalid command format."
+              "Please use 'mv source_file destination_path'.")
+    try:
+        os.rename(source_file, destination_file)
+    except Exception as e:
+        print(e)
```
Dimosphen1, line 29: You should not rely on the filesystem delimiters such as `"/"` in this solution, your code should be platform-agnostic. Refactor your code to encounter that.

Dimosphen1 (changes requested):

PR #494, opened 2023-12-15
all tests passed

bkirichenko:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3:
+        return
```
vsmutok, line 7: I think better to raise some informative error

app/main.py
```diff
@@ -1 +1,18 @@
...
+
+    command, source, target = parts
+    if any((command != "mv", source == target)):
+        return
```
vsmutok, line 11: I think better to raise some informative error

vsmutok: so that the user knows if something goes wrong

vsmutok (changes requested):

PR #501, opened 2023-12-28
Solution

ShcherbynaSv:


app/main.py
```diff
@@ -1 +1,34 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command, in_file, out_file = command.split()
```
Dimosphen1, line 5: To avoid an error related to unpacking a list that contains more than 3 elements after split, you could add the following syntax
```suggestion
    command, in_file, out_file, *_ = command.split()
```

app/main.py
```diff
@@ -1 +1,34 @@
...
+    if command == "mv":
+        path = out_file
+        path_list = []
+        while path != "":
+            path_list.append(os.path.split(path)[1])
+            path = os.path.split(path)[0]
```
Dimosphen1, line 11: You could obtain a path more simply without iterations or indexing, e.g.
```suggestion
        path = os.path.normpath(out_file)
        path = path.split(os.sep)
        *file_path, file_name = path
        
```
Rewrite code using this approach to avoid indexing and iterations through the path.

app/main.py
```diff
@@ -1 +1,34 @@
...
+        if len(path_list) == 1:
+            with (open(in_file, "r")as start_file,
+                  open(out_file, "w") as end_file):
+                end_file.write(start_file.read())
+
+        else:
+            path_list = path_list[3::-1]
+
+            path = ""
+            for i in range(len(path_list) - 1):
```
Dimosphen1, line 22: Do not use single-letter variable names, choose more descriptive naming.

app/main.py
```diff
@@ -1 +1,34 @@
...
+                if os.path.exists(os.path.join(path, path_list[i])):
+                    path = os.path.join(path, path_list[i])
+                    print(path)
+                else:
+                    os.mkdir(os.path.join(path, path_list[i]))
+                    path = os.path.join(path, path_list[i])
+                    print(path)
+            with (open(in_file, "r")as start_file,
```
Dimosphen1, line 30:
```suggestion
            with (open(in_file, "r") as start_file,
```

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,33 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command, in_file, out_file, *_ = command.split()
```
vasylhnatiuk, line 5: return None if command.split() has less then 3 arguments

app/main.py
```diff
@@ -1 +1,33 @@
...
+    if command == "mv":
+        # path = out_file
+        path = os.path.normpath(out_file)
+        print(path)
```
vasylhnatiuk, line 9: delete uselesss prints

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,30 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if len(command.split()) < 3:
```
vasylhnatiuk, line 5:
```suggestion
    if len(command) < 3:
```

vasylhnatiuk: it aslo will be working

vasylhnatiuk (approved):

PR #504, opened 2023-12-28
Solution

KwintDimon:


app/main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    cd, source_name, target_name, *_ = command.split()
```
vasylhnatiuk, line 5: what if command have less then 3 arguments? add check before

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,33 @@
...
+        return
+    cd, source_name, target_name, *_ = command_parts
+    if cd != "mv":
+        return
+
+    if "/" in target_name:
+        *target_path_parts, file_name = target_name.split("/")
+        current_path = ""
+        for directory in target_path_parts:
+            current_path = os.path.join(current_path, directory)\
```
vasylhnatiuk, line 17:
```suggestion
            current_path = os.path.join(current_path, directory)
```

vasylhnatiuk (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #510, opened 2024-01-07
'Solution' py-move-file

yura-push:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_name, old_file, path = command.split()
```
Dimosphen1, line 5: You could consider adding `*_` to collect excessive items in the command after split
```suggestion
    command_name, old_file, path, *_ = command.split()
```

app/main.py
```diff
@@ -1 +1,18 @@
...
+    path_dirs = path.split("/")
```
Dimosphen1, line 6: Do not use any specific directory delimiter in your code, your code should be platform-agnostic.

app/main.py
```diff
@@ -1 +1,18 @@
...
+    new_file = path_dirs.pop(-1)
+
+    with open(old_file, "r") as file_out:
+        content = file_out.read()
+        new_path = ""
+        for directory in path_dirs:
+            new_path += directory + "/"
+            if not os.path.exists(new_path):
+                os.mkdir(new_path)
```
Dimosphen1, line 15: You could replace this part with `os.makedirs(path_dirs, exist_ok=True)`

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_name, old_file, path = command.split()
+    path_dirs = path.split("/")
+    new_file = path_dirs.pop(-1)
+
+    with open(old_file, "r") as file_out:
```
Dimosphen1, line 9: Instead of opening one context manager inside of another, you could open two content managers at once.

app/main.py
Dimosphen1, line 5: You also need to check that the command is `mv` and two afterward arguments are not the same.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_name, old_file, path, *_ = command.split()
+
+    if command_name == "mv" and old_file != path:
+        last_slash_index = path.rfind("/")
```
Dimosphen1, line 8: Do not use any specific directory delimiter in your code, your code should be platform-agnostic. Use methods from the `os` module to obtain the path and filename.

app/main.py
```diff
@@ -1 +1,14 @@
...
+        path_dirs = path[:last_slash_index]
+        os.makedirs(path_dirs, exist_ok=True)
+        with open(old_file, "r") as file_out, open(path, "w") as file_in:
+            content = file_out.read()
+            file_in.write(content)
```
Dimosphen1, line 13: There is no need to allocate a `content` variable, you could omit it
```suggestion
            file_in.write(file_out.read())
```

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #519, opened 2024-01-11
Solution

maxbidun:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    mv, file_in, file_out = command.split()
```
vsmutok, line 5: you must check before unpacking that the length of `command.split()` is three

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if len(command.split()) == 3 and command.split()[1] != command.split()[2]:
+        mv, file_in, file_out = command.split()
```
vsmutok, line 6: DRY - don't repeat yourself 

`command.split()` - you only have to do it once, now you do it 4 times

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #520, opened 2024-01-12
Solution

msymonovych:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if len(command.split()) != 3:
+        return
+    operation, source_file, new_file = command.split()
```
NEkropo1, line 7: you splitting command twice, better make variable

app/main.py
```diff
@@ -1 +1,18 @@
...
+    if operation != "mv":
+        return
```
NEkropo1, line 9: also try to refactor function to make no returns, for example, here you can check `if opertaion == "mv"`

app/main.py
```diff
@@ -1 +1,18 @@
...
+    dirs = os.path.dirname(new_file)
+    if dirs:
+        os.makedirs(dirs, exist_ok=True)
+
+    with open(source_file, "r") as source:
+        with open(new_file, "w") as target:
+            target.writelines(source.read())
+
+    os.remove(source_file)
```
NEkropo1, line 18: this is ok, but you can use more pythonic way (import replace, for example)

NEkropo1 (approved):
Good job!

PR #524, opened 2024-01-15
Solution

Zonda001:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+import os
+import shutil
+
+
+def move_file(command: str) -> None:
+    parts = command.split()
+    if not parts or parts[0] != "mv" or len(parts) != 3:
+        return
```
vsmutok, line 8: it is better to raise some exception so that it is clear to the user if something goes wrong

app/main.py
```diff
@@ -1 +1,20 @@
...
+    _, source_file, destination = parts
+    if source_file == destination:
+        return
```
vsmutok, line 11: it is better to raise some exception so that it is clear to the user if something goes wrong

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #525, opened 2024-01-16
Solution

Vasavik-ua:


app/main.py
```diff
@@ -1 +1,15 @@
...
+def move_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) == 3 and parts[0] == "mv":
+        if parts[1] == parts[2]:
+            return
+        directory, _ = os.path.split(parts[2])
+        if len(directory) != 0:
+            os.makedirs(directory, exist_ok=True)
+        with (open(parts[1], "r") as file_in,
+              open(parts[2], "w") as file_out):
```
vsmutok, line 13: it's better to unpack values into the variables, like: var, var, var = parts. Then use these variables instead of indexes, it will increase readability

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if len(command.split()) == 3:
+        command, file_name, path = command.split()
```
vsmutok, line 6: DRY - do not use `command.split()` twice

app/main.py
```diff
@@ -1 +1,16 @@
...
+        if file_name == path:
+            return
```
vsmutok, line 8: it is better to raise some exception so that it is clear to the user if something goes wrong

vsmutok (changes requested):

app/main.py
Vasavik-ua: what is mean DRY??

app/main.py
vsmutok: do it

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    try:
+        command, file_name, path = command.split()
+        if file_name == path:
+            return
```
vsmutok, line 8:
```suggestion
    parts = command.split()
    if len(parts) != 3:
        raise raise Exception(....) 
```

vsmutok (changes requested):

PR #531, opened 2024-01-19
Implement move-file task

Dmytry-S:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    com_d = command.split()
```
NazarSheremet, line 5: It's better to assign each part of the command to separate variables, making them easier to understand when used.
```suggestion
    command, source_path, destination_path = command.split()
```

app/main.py
```diff
@@ -1 +1,15 @@
...
+    if com_d[0] != "mv":
+        return None
+    n_dir = com_d[2].split("/")
```
NazarSheremet, line 8: Better to use `os.path.split()`
```suggestion
    directory, filename = os.path.split(com_d[2])
```

app/main.py
```diff
@@ -1 +1,15 @@
...
+    if len(n_dir) == 1:
+        os.rename(com_d[1], com_d[2])
+    else:
+        dir_path = "/".join(n_dir[:-1]) + "/"
+        dir_path = os.path.join(dir_path, os.path.basename(n_dir[-1]))
```
NazarSheremet, line 13: If you implement my previous suggestions this code will not be needed

NazarSheremet (changes requested):
Nice

Dmytry-S:
Changed, have a look.

NazarSheremet (approved):
Good Job!

PR #533, opened 2024-01-20
Solution

IvanGuculyak:


app/main.py
```diff
@@ -1 +1,26 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if len(command.split()) != 3:
+        return
```
vsmutok, line 6: it is better to raise informative error so that it is clear to the user if something goes wrong

app/main.py
```diff
@@ -1 +1,26 @@
...
+
+    operation, file1, path_and_file2 = command.split()
+    path = ""
+
+    if "/" in path_and_file2:
+        path = "/".join(path_and_file2.split("/")[:-1]) + "/"
+        file_name = path + path_and_file2.split("/")[-1]
+    else:
+        file_name = path_and_file2
```
vsmutok, line 15: use "os.path.dirname" instead

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,26 @@
...
+    path = ""
+
+    if "/" in path_and_file2:
+        path = os.path.dirname(path_and_file2)
+        file_name = path + "/" + path_and_file2.split("/")[-1]
+    else:
+        file_name = path_and_file2
+
+    if len(path) != 0:
+        os.makedirs(path, exist_ok=True)
```
vsmutok, line 18:
```suggestion
    path = os.path.dirname(new_path)
    if path:
        os.makedirs(path, exist_ok=True)
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,23 @@
...
+    if path:
+        os.makedirs(path, exist_ok=True)
+        file_name = path + "/" + path_and_file2.split("/")[-1]
+    else:
+        file_name = path_and_file2
+
+    with (
+        open(file1, "r") as source,
+        open(file_name, "w") as new_file
+    ):
```
vsmutok, line 20:
```suggestion
      path = os.path.dirname(new_path)
      if path:
          os.makedirs(path, exist_ok=True)

    with (
        open(file1, "r") as source,
        open(path_and_file2, "w") as new_file
    ):
```

vsmutok: also `file1 ` not clear name

vsmutok: Make sure you check that the mv input command is correct.

vsmutok (changes requested):

vsmutok (approved):
Good Job!

PR #536, opened 2024-01-24
Solution

Disash:


app/main.py
```diff
@@ -1 +1,20 @@
...
+from os import makedirs, path, remove, replace
+
+
+class ErrorCommand(Exception):
+    pass
+
+
+def move_file(command: str) -> None:
+    if len(command.split()) == 3:
+        mv, source_file, destination_file = command.split()
```
EdAlekseiev, line 10: Don't use `command.split()` twice. Create a separate variable for storing result of `command.split()` and use it where you need.

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #537, opened 2024-01-24
'Solution'

OLENA-KALITSINSKA:


app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    split_command = command.split()
+    source_path = split_command[1]
+    destination_path = split_command[2]
+    if (len(split_command) == 3
```
pavlejviki, line 8: You should check the length before indexing, to make sure the index exists

app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    split_command = command.split()
+    source_path = split_command[1]
+    destination_path = split_command[2]
```
pavlejviki, line 7: use unpacking better

pavlejviki (changes requested):

PR #538, opened 2024-01-24
'Solution'

JohnnyWalker010:


barabarinov (approved):
GJ!

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    operation, source_file, new_file = command.split()
+
```
NEkropo1, line 6: what if operation is wrong?

NEkropo1 (changes requested):

NEkropo1 (approved):
Good job!

PR #544, opened 2024-02-01
Devops

volry:


volry:
Solution

YuriiSmolii (changes requested):
Занадто багато коду та перевірок для цього завдання, перепиши функцію, подумай над тим як можна скоротити її

app/main.py
```diff
@@ -1 +1,41 @@
...
+import shutil
+
+
+def move_file(command: str) -> None:
+    parts = command.split(" ")
+    if len(parts) != 3 or parts[0] != "mv":
+        raise ValueError("Invalid command")
+
+    source_file = parts[1]
+    destination = " ".join(parts[2:])
```
NazarSheremet, line 11: You can do this in `split()` function 
```suggestion
     cmd, source_file, destination = command.split()
```

app/main.py
```diff
@@ -1 +1,41 @@
...
+
+    if destination.startswith("dir") and os.path.exists("dir"):
+        shutil.rmtree("dir")
```
NazarSheremet, line 14: Sorry but I don't understand what this condition check and why need to remove "dir" please add a comment with explanation

app/main.py
```diff
@@ -1 +1,41 @@
...
+        or (os.path.isdir(destination) and not os.path.isfile(destination))
+    ):
+        os.makedirs(destination, exist_ok=True)
+        dest = os.path.join(destination, os.path.basename(source_file))
+    else:
+        # Ensure the destination's parent directory exists
+        parent_dir = os.path.dirname(destination)
+        if parent_dir and not os.path.exists(parent_dir):
+            os.makedirs(parent_dir, exist_ok=True)
+        dest = destination
```
NazarSheremet, line 27: Simplify this code. Add the first condition to check if the destination is the same as source and after chech if directory exist, if not create it

NazarSheremet (changes requested):
Need some changes

app/main.py
```diff
@@ -1 +1,40 @@
...
+
+def move_file(command: str) -> None:
+    cmd, source_file, des = command.split()
+
+    if cmd != "mv" or len([cmd, source_file, des]) != 3:
+        raise ValueError("Invalid command")
+
+    # Check if the destination starts with "dir" and if "dir" exists, remove it
+    if des.startswith("dir") and os.path.exists("dir"):
+        shutil.rmtree("dir")
```
NazarSheremet, line 13: Don't understand, why this check is needed

NazarSheremet (changes requested):
Need fix or explanation

app/main.py
volry: you are right, it's not needed.

NazarSheremet (approved):
Good job!

PR #553, opened 2024-02-04
Solution

AlexGrytsai:


app/main.py
```diff
@@ -1 +1,32 @@
...
+
+def move_file(command: str) -> None:
+    command_list = command.split()
+    if len(command_list) != 3:
+        raise ValueError(f"Command must have 3 elements, but got: "
+                         f"{len(command_list)}")
+    move_command, old_file_path, new_file_path = command_list
+    if move_command != "mv":
+        raise NameError(f"{move_command} is not a valid")
+    if new_file_path.count("/") == 0:
```
vasylhnatiuk, line 13: Use os.path.join to concatenate parts of the path. In that case, your app will be cross-platform and will use either / or \ depending on OS

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -10,14 +11,19 @@ def move_file(command: str) -> None:
     move_command, old_file_path, new_file_path = command_list
     if move_command != "mv":
         raise NameError(f"{move_command} is not a valid")
-    if new_file_path.count("/") == 0:
+    try:
+        new_file_path_components = re.split(r"[/\\]", new_file_path)
```
vasylhnatiuk, line 15: use os.path.split()

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #554, opened 2024-02-05
solution

MilArtem78:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command = command.split()
```
Dimosphen1, line 5: You could perform unpacking during the split operation, in this way, you will avoid using indexing and make your code more descriptive:
```suggestion
    command_name, source_file, destination, *_ = command.split()
```

app/main.py
```diff
@@ -1 +1,12 @@
...
+    if len(command) == 3 and command[0] == "mv":
+        source_file, destination = command[1], command[2]
+        end_path, end_file = os.path.split(destination)
+        if end_path != "":
+            if not os.path.exists(end_path):
+                os.makedirs(end_path)
```
Dimosphen1, line 11: There is no need to create this if condition, call `makedirs` with `exist_ok` flag instead.

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #555, opened 2024-02-06
Solution

Damodar-kumar:


app/main.py
```diff
@@ -1 +1,21 @@
...
+
+def move_file(command: str) -> None:
+    words_of_command = command.split(" ")
+    if (len(words_of_command) == 3
+        and words_of_command[0] == "mv"
+            and words_of_command[1] != words_of_command[2]):
+        if words_of_command[2].count("/") == 0:
+            os.rename(words_of_command[1], words_of_command[2])
+        else:
+            create_directory = words_of_command[2].split("/")
```
vasylhnatiuk, line 12: Use os.path.join to concatenate parts of the path. In that case, your app will be cross-platform and will use either / or \ depending on OS. And also you should use os.path.split()

vasylhnatiuk (changes requested):

PR #557, opened 2024-02-08
solution

yaaszp:


app/main.py
```diff
@@ -1 +1,24 @@
...
+    os.remove(list_from_input_str[1])
+    if "/" in list_from_input_str[2]:
+        list_dir_and_file = list_from_input_str[2].split("/")
+        output_file_name = list_dir_and_file[-1]
+        for i in range(len(list_dir_and_file) - 1):
+            if os.path.exists(list_dir_and_file[i]):
+                os.chdir(list_dir_and_file[i])
+            else:
+                os.mkdir(list_dir_and_file[i])
+                os.chdir(list_dir_and_file[i])
```
NazarSheremet, line 18: This solution looks like a few overheads, You can get a full path to the file using  `os.path.dirname()` and after use os.makedirs() to create all dirs by one command

NazarSheremet (changes requested):
Need to optimaze

app/main.py
yaaszp: fixed

NazarSheremet (approved):
Good job!

PR #558, opened 2024-02-09
Solution

MykhailoIvchenko:


app/main.py
```diff
@@ -1 +1,42 @@
...
+        os.rename(old_file_path, new_file_path)
+        return
+
+    new_file_name = new_file_path[slash_last_index:]
+
+    dirs_path = new_file_path[0:slash_last_index]
+
+    cur_path = ""
+
+    directories = dirs_path.split("/")
```
EdAlekseiev, line 25: You should use `os.path.split`.

app/main.py
```diff
@@ -1 +1,42 @@
...
+
+    for directory in directories:
+        cur_path += (directory + "/")
+
+        if not os.path.isdir(cur_path):
+            os.mkdir(cur_path)
```
EdAlekseiev, line 31: Use os.makedirs with exist_ok = False

EdAlekseiev (changes requested):

app/main.py
MykhailoIvchenko: Done

app/main.py
```diff
@@ -1 +1,32 @@
...
+        return
+
+    new_file_name = new_path_parts[1]
+
+    os.makedirs(new_path_parts[0], exist_ok=True)
+
+    with open(old_file_path, "r") as file:
+        content = file.read()
+
+    path_to_move = new_file_path if new_file_name else (new_file_path
```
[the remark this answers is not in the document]
MykhailoIvchenko: Fixed. But this code wouldn't work for this case: "If name in destination path ends with / it must be considered as a directory", because here we don't have the file name in the destination path. There is no test for this case so tests are passed successfully. But there is such case in the task.

app/main.py
```diff
@@ -1 +1,29 @@
...
+    if not new_path_parts[0]:
+        os.rename(old_file_path, new_file_path)
+        return
+
+    os.makedirs(new_path_parts[0], exist_ok=True)
+
+    with open(old_file_path, "r") as file:
+        content = file.read()
+
+    path_to_move = new_file_path
```
YukihiroSM, line 24: Agree that this doesn't cover the case with / in the end of destination path, but it should be something different from yours solution: 
`path_to_move = new_file_path if os.path.isfile(new_path_parts[-1]) else (new_file_path + old_path_parts[-1])`
Try to improve this line of code for you purposes.

YukihiroSM (changes requested):

app/main.py
MykhailoIvchenko: The os.path.isfile() is unsuitable in the discussed case because we make a check before the file creation, so this expression's result is always False.  I've used the new_path_parts list, but without the isfile() method. I also have added the test for the case, when the target file name is not provided (the slash in the destination path at the end).

vasylhnatiuk (approved):

PR #568, opened 2024-02-18
solution

Eugen-Veklich:


app/main.py
```diff
@@ -1 +1,19 @@
...
+
+def move_file(command: str) -> None:
+    commands = command.split(" ")
+    destination = commands[2].split("/")
+    new_file = destination.pop()
+    target_directory = ""
+    if len(commands) == 3:
+        if commands[0] == "mv":
+            for folder in destination:
+                target_directory += folder + "/"
```
EdAlekseiev, line 12: Use os.path.join

app/main.py
```diff
@@ -1 +1,19 @@
...
+                if os.path.isdir(target_directory):
+                    continue
+                os.mkdir(target_directory)
```
EdAlekseiev, line 15: use os.makedirs with exist_ok=True to avoid OSError

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #572, opened 2024-02-19
Develop

kolya2731:


app/main.py
```diff
@@ -1 +1,11 @@
...
+
+
+def move_file(command: str) -> None:
+    ls_of_command = command.split()
+
+    destination_dir = os.path.dirname(ls_of_command[-1])
+    if destination_dir and not os.path.exists(destination_dir):
+        os.makedirs(destination_dir)
+
+    os.rename(ls_of_command[1], ls_of_command[-1])
```
Dimosphen1, line 11: - You should make sure that the first command is `mv` as expected;
- Use unpacking and perform a length check before doing that;
- Use the `exist_ok=True` flag to avoid checking the directory presence;
- Do not use indexing to make code more clear, use variables obtained during unpacking.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #573, opened 2024-02-21
SOL

neagarr:


neagarr:
I wasn’t able to use your option because the logic didn’t work correctly, but I managed to make this code more readable. I would be grateful if you could explain with an example how I should have used the line of code you suggested.

PR #578, opened 2024-02-28
Solution

RomanNest:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    mv, file, path_to = command.split()
+
+    if mv == "mv" and len(command.split()) == 3 and file != path_to:
```
EdAlekseiev, line 7: Don't use `split` twice. You can create varable for storring result of `split` and use it where you need.

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #580, opened 2024-03-01
Solution

twonumberfortyfives:


app/main.py
```diff
@@ -1 +1,17 @@
...
+def move_file(command: str) -> None:
+    commands = command.split()
+    input_file = commands[1]
+    output_file = commands[2]
+    with open(input_file, "r") as i_file:
+        input_content = i_file.read()
+    if "/" in output_file:
+        output_directory = os.path.dirname(output_file)
+        os.makedirs(output_directory, exist_ok=True)
+        with open(output_file, "w") as o_file:
```
vasylhnatiuk, line 13: try to use `with` only once

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    commands = command.split()
+    input_file = commands[1]
```
vasylhnatiuk, line 6: you should check if command is correct and has 3 arguments

app/main.py
```diff
@@ -1 +1,17 @@
...
+    output_file = commands[2]
+    with open(input_file, "r") as i_file:
```
vasylhnatiuk, line 8: i_file is not an appropriate name

vasylhnatiuk (changes requested):

PR #583, opened 2024-03-04
Solution

mish-sk:


app/main.py
```diff
@@ -1 +1,24 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    split_command = command.split(" ")
```
EdAlekseiev, line 5: `" "` is default value for split
```suggestion
    split_command = command.split()
```

app/main.py
```diff
@@ -1 +1,24 @@
...
+    if split_command[0] != "mv":
```
EdAlekseiev, line 6: Also, check that `split_command` contains 3 elements

app/main.py
```diff
@@ -1 +1,24 @@
...
+    if original_file == new_file_path:
+        return
+
+    directory, filename = os.path.split(new_file_path)
+
+    if directory:
+        os.makedirs(directory, exist_ok=True)
+
+    with open(original_file, "r") as old_file:
+        with open(split_command[2], "w") as new_file:
```
EdAlekseiev, line 21: Use one `with`

EdAlekseiev (changes requested):

PR #592, opened 2024-03-13
Solution

MrsLFG:


app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+import os
+import shutil
+
+
+def move_file(command: str) -> None:
+    parts = command.split()
+
```
EdAlekseiev, line 7: You should check, that `parts` contains 3 elements and first part of command is `mv`

EdAlekseiev (changes requested):

PR #603, opened 2024-03-19
Solution

Bujhm99:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+import os
+import shutil
+
+
+def move_file(command: str) -> None:
+    command_list = command.split(" ")
```
nicksetrakov, line 6: you could write just .split()

nicksetrakov (changes requested):

app/main.py
```diff
@@ -1 +1,15 @@
...
+
+
+def move_file(command: str) -> None:
+    command_list = command.split()
+    command_list[2] = command_list[2].split("/")
+    new_dir = ""
+
+    for index in range(len(command_list[2]) - 1):
+        if not os.path.exists(new_dir + command_list[2][index]):
+            os.mkdir(new_dir + command_list[2][index])
```
YuriiKindrat, line 12: try to use mkdirs(exists_ok=True) intead here)

YuriiKindrat (changes requested):
u are using to much indexes in ur code, this make this code really hard to read
try to remove indexes from ur code

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+import os
+import shutil
+
+
+def move_file(command: str) -> None:
+    _, current_file, new_file_path = command.split()
```
Oleksl888, line 6: What if user passes command wrong (for example uses wrong command or passes 2 or 4 args instead of 3 that you expect?)

app/main.py
```diff
@@ -1 +1,15 @@
...
+    dir_list = new_file_path.split("/")
```
Oleksl888, line 7: `os.path.split()` will be better here since it will correctly split on all systems

app/main.py
```diff
@@ -1 +1,15 @@
...
+    new_file_name = dir_list.pop()
+    new_dir = ""
+
+    for directory in dir_list:
+        os.makedirs(new_dir + directory, exist_ok=True)
```
Oleksl888, line 12: better use `os.path.join` here so it works correctly on all systems

app/main.py
```diff
@@ -1 +1,15 @@
...
+import shutil
+
+
+def move_file(command: str) -> None:
+    _, current_file, new_file_path = command.split()
+    dir_list = new_file_path.split("/")
+    new_file_name = dir_list.pop()
+    new_dir = ""
+
+    for directory in dir_list:
```
Oleksl888, line 11: you don't really need a for loop here since `makedirs` will create all necessary dirs included in destination path if they don't exist

app/main.py
```diff
@@ -1 +1,15 @@
...
+        os.makedirs(new_dir + directory, exist_ok=True)
+        new_dir += directory + "/"
+    shutil.copy(current_file, new_dir + new_file_name)
+    os.remove(current_file)
```
Oleksl888, line 15: when you work with files it's better to use safety nets like check if file exists before moving/deleting it or use exceptions. Also these two command can be replaced with `shutil.move`

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,64 @@
-# write your code here
+import os
+import shutil
+
+
+class Errors(BaseException):
```
Oleksl888, line 5: Your own exceptions shouldn't be inherited from BaseException - they are too general. Don't overcomplicate this,  OsError will cover most cases

app/main.py
```diff
@@ -1 +1,64 @@
...
+class IsNotFileEr(Errors):
+    pass
+
+
+class NoSuchFile(Errors):
+    pass
+
+
+def is_smth_wrong_command_right(command: str) -> None:
+    if not isinstance(command, str):
```
Oleksl888, line 30: This is too much spaghetti. Your input validation should be like:
1. Check if you received exactly 3 arguments.
2. Check if comand is 'mv'.
3. After this you can safely split your command into 3 pieces.
4. Then you use `os.path.join/split` methods to get destination path.
5. Use `os.makedirs` to create dirs and `shutil.move` for main logic - wrap them in exceptions to handle wrong dir/file names and handle non-exsiting files.
6. This whole thing shouldn't take you more then 15 lines of code to write

Oleksl888 (changes requested):

YuriiKindrat (approved):
GJ, but it is really hard to read it with indexes (

PR #604, opened 2024-03-20
Solution?

OKochubeii:


app/main.py
```diff
@@ -1 +1,17 @@
...
+import os.path
+
+
+def move_file(command: str) -> None:
+    command_split = command.split(" ")
+    if command_split[0] != "mv":
+        return
+
+    file_orig = command_split[1]
+    file_new = command_split[2]
```
NazarSheremet, line 10: It can be done simpler, but also need to add a check if the command contains 3 parts
`command_name, file_orig, file_new = command.split()`

NazarSheremet (changes requested):

app/main.py
OKochubeii: Дякую за відгук, вже виправив🫡

NazarSheremet (approved):
Good job!

PR #607, opened 2024-03-21
Pochtalon

pochtalon:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    mv, old_file, new_file = command.split(" ")
+    if old_file == new_file:
+        return
+    path = new_file.split("/")
+    new_dir = "/".join(path[:-1])
```
NazarSheremet, line 9: Will be better to use the function  `os.path.dirname()` to separate dir name from all path

NazarSheremet (changes requested):
Minor changes

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    mv, old_file, new_file = command.split(" ")
+    if old_file == new_file:
+        return
+    new_dir = os.path.dirname(new_file)
+    if new_dir != "":
```
NazarSheremet, line 9: In Python, you can use a simple condition to check if a string is not empty.
```suggestion
    if new_dir:
```

NazarSheremet (changes requested):
Sorry, I missed this incorrection during the first check.

NazarSheremet (approved):
Good job!

PR #613, opened 2024-03-25
Solution

Luxiv:


app/main.py
```diff
@@ -1 +1,23 @@
...
+
+
+def move_file(command: str) -> None:
+
+    source, destination = command.split()[1:]
+
+    source_parts = source.rsplit("/", 1)
+    destination_parts = destination.rsplit("/", 1)
+
+    if len(destination_parts) > 1:
```
YuriiSmolii, line 11: You are using rsplit with max splits 1 so this condition will never match.
Chech your function locally.

YuriiSmolii (changes requested):
Refactor logic.

PR #617, opened 2024-03-26
Solution

Line181:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_list = command.split()
+    if len(command_list) == 3:
```
pavlopro, line 6: Also check that first element is 'mv'

pavlopro: Would be good to simplify this code

pavlopro (changes requested):

PR #619, opened 2024-03-27
First_try

Sydorovich:
used only functions in readme of the repo + os.path.exists

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,20 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command = command.split()
+    initial_file, destination = command[1], command[2]
+    if "/" not in command[2]:
+        os.rename(initial_file, destination)
+        return
+    with open(initial_file, "r") as file_in:
+        content = file_in.read()
+    os.remove(initial_file)
+    path = destination.split("/")[:-1]
+    step = ""
+    for i in path:
+        step += i + "/"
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command = command.split()
+    initial_file, destination = command[1], command[2]
```
NazarSheremet, line 6: You can do this like: 
```suggestion
    command, initial_file, destination = command.split()
```

app/main.py
```diff
@@ -1 +1,20 @@
...
+    if "/" not in command[2]:
+        os.rename(initial_file, destination)
+        return
+    with open(initial_file, "r") as file_in:
+        content = file_in.read()
+    os.remove(initial_file)
+    path = destination.split("/")[:-1]
```
NazarSheremet, line 13: You can use `os.path.dirname()` functions to separate path from file name

app/main.py
```diff
@@ -1 +1,20 @@
...
+    step = ""
+    for i in path:
+        step += i + "/"
+        if os.path.exists(step) is False:
+            os.mkdir(step)
```
NazarSheremet, line 18: `os.makedirs()`  function allows you to create the entire sequence of folders along the path.

NazarSheremet (changes requested):
Need changes

PR #622, opened 2024-03-29
Solution

vitalinamal:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    _, source_file, destination_path = command.split()
+    if "/" in destination_path:
+        os.makedirs(os.path.dirname(destination_path), exist_ok=True)
+    with open(source_file, "r") as src, open(destination_path, "w") as dst:
+        dst.write(src.read())
```
panicua, line 9: use descriptive names, for example: ```src``` = ```source_file```, ```dst``` = ```destination_file```

panicua (changes requested):

panicua (approved):
GJ. 
I actually stole os.makedirs() method from your code after I reviewed it cause I were using ```os.mkdir()``` in a for loop and had to use ```try/except``` because it was giving an error if folder is already exists and it doesn't have the ```exist_ok``` argument to handle it.

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    _, source_file, destination_path = command.split()
+    if "/" in destination_path:
```
Oleksl888, line 6: What if path separators are `\`?

app/main.py
```diff
@@ -1 +1,11 @@
...
+        os.makedirs(os.path.dirname(destination_path), exist_ok=True)
```
Oleksl888, line 7: Directories cannot be created if you use certain symbols in names. Does your code handle this?

app/main.py
```diff
@@ -1 +1,11 @@
...
+    with (open(source_file, "r") as copied_file,
```
Oleksl888, line 8: There is a function in `shutil` library that does all this in one line. Also how does this code handle cases when input file does not exist?

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,21 @@
...
+import os
+import shutil
+
+
+def move_file(command: str) -> None:
+    source_file = None
+    try:
+        _, source_file, destination_path = command.split()
+
+        if any(char in source_file + destination_path for char in '<>:"|?*'):
```
Oleksl888, line 10: this is unneccessary - if you put `os.makedirs` and `shutil.move` in try-except clauses an exception will be triggered that you are already handling

app/main.py
```diff
@@ -1 +1,21 @@
...
+            raise ValueError("file name contains invalid characters")
+        if "/" in destination_path or "\\" in destination_path:
```
Oleksl888, line 12: better use `os.path.splt` method

app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+import os
+import shutil
+
+
+def move_file(command: str) -> None:
+    source_file = None
+    try:
+        _, source_file, destination_path = command.split()
```
Oleksl888, line 8: `Make sure you check that the mv input command is correct.`
this requirement is not satisifed

app/main.py
```diff
@@ -1 +1,21 @@
...
+
+        if any(char in source_file + destination_path for char in '<>:"|?*'):
+            raise ValueError("file name contains invalid characters")
+        if "/" in destination_path or "\\" in destination_path:
+            os.makedirs(os.path.dirname(destination_path), exist_ok=True)
+        shutil.move(source_file, destination_path)
+
+    except FileNotFoundError:
```
Oleksl888, line 16: Too many Exceptions - you can use ValueError to track errors related to line where you split() the command. And all file related errors will be covered by OSError.
Also make sure you include descriptive messages if you raise exceptions

Oleksl888 (changes requested):

YuriiKindrat (approved):
GJ

PR #624, opened 2024-03-29
Solution

panicua:


panicua:
removed ```for``` loop, ```try/except``` after first commit. Instead used ```os.makedirs()``` method.

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    _, start_file_name, destination_path = command.split(" ")
```
Oleksl888, line 5: No need to `split(" ")` explicitly split() splits by whitespace by default

app/main.py
```diff
@@ -1 +1,11 @@
...
+    destination_path_list = destination_path.split("/")
```
Oleksl888, line 6: you should use `os.path` methods to correctly split on all systems.
Same with join method below

app/main.py
Oleksl888, line 5: `Make sure you check that the mv input command is correct.` 
what about this?

app/main.py
```diff
@@ -1 +1,11 @@
...
+    if not len(destination_path_list) == 1:
+        path = "/".join(destination_path_list[:len(destination_path_list) - 1])
```
Oleksl888, line 8: What if destination_path_list contains unacceptable symbols?

app/main.py
```diff
@@ -1 +1,11 @@
...
+        os.makedirs(path, exist_ok=True)
+
+    os.rename(start_file_name, destination_path)
```
Oleksl888, line 11: What if filename is wrong or file not exist?

Oleksl888 (changes requested):

panicua:
I forgot to look at checklist before push, same as with the last one task

fsocie7y (approved):
well done ^_^

PR #625, opened 2024-03-31
Solution

DmytroHlazyrin:


panicua (approved):
GJ

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    move, file_in, file_out = command.split(" ")
```
Oleksl888, line 5: you don't need to pass double quotes in split method

app/main.py
```diff
@@ -1 +1,18 @@
...
+    if folder_in_path == folder_out_path:
+        os.rename(file_in, file_out)
+    else:
+        if not os.path.exists(folder_out_path):
+            os.makedirs(folder_out_path)
+
+        with (open(file_in, "r") as file_in,
+              open(file_out, "w") as file_out):
+            file_out.write(file_in.read())
+        os.remove(file_in.name)
```
Oleksl888, line 18: you can do this in one line with function from `shutil` library

app/main.py
Oleksl888, line 5: You need to check if `mv` is the first token in the command as per task requirements

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    move, file_in, file_out = command.split(" ")
+    folder_out_path = os.path.dirname(file_out)
+    folder_in_path = os.path.dirname(file_in)
```
Oleksl888, line 7: How does your code handle input when source file or destination have unacceptable symbols

Oleksl888 (changes requested):

YuriiKindrat (approved):
LGTM

PR #626, opened 2024-04-01
Solution

D-u-x:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    if len(command.split()) == 3 and command.split()[0] == "mv":
+        *_, in_file, out_file = command.split(" ")
```
EdAlekseiev, line 6: Don't call command.split() a few times. You should create separate variable for storing result of command.split() and use it where you need

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_list = command.split()
+
+    if len(command_list) == 3 and command_list[0] == "mv":
+        *_, in_file, out_file = command.split(" ")
```
EdAlekseiev, line 8:
```suggestion
        *_, in_file, out_file = command_list
```

EdAlekseiev (changes requested):

vasylhnatiuk (approved):

PR #628, opened 2024-04-01
Solution for task move file

oleksandra-shershen:


app/main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+import os
+import shutil
+
+
+def move_file(command: str) -> None:
+    command_parts = command.split()
+    command_name, source_file, destination_file = command_parts
```
vitalinamal, line 7: I would write a length check immediately after the split and than unpacking

vitalinamal (approved):
Good

app/main.py
```diff
@@ -1 +1,29 @@
...
+        raise FileNotFoundError(f"Source file not found: {source_file}")
+
+    source_directory = os.path.dirname(source_file)
+    destination_directory = os.path.dirname(destination_file)
+
+    if source_directory == destination_directory:
+        os.rename(source_file, destination_file)
+    else:
+        if not os.path.exists(destination_directory):
+            os.makedirs(destination_directory)
```
Oleksl888, line 22: you can pass `exist_ok=True` so it doesn't trigger exception if the dir already exists

Oleksl888 (changes requested):

app/main.py
oleksandra-shershen: Resolved

app/main.py
```diff
@@ -1 +1,29 @@
...
+    if len(command_parts) != 3:
+        raise ValueError("Invalid move command format")
+
+    if not os.path.exists(source_file):
+        raise FileNotFoundError(f"Source file not found: {source_file}")
+
+    source_directory = os.path.dirname(source_file)
+    destination_directory = os.path.dirname(destination_file)
+
+    if source_directory == destination_directory:
```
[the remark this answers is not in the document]
oleksandra-shershen: Resolved

app/main.py
oleksandra-shershen: Resolved

LiudmylaKulinchenko (approved):
Well done 🌞

PR #631, opened 2024-04-05
'solution'

liudmylaperchevska:


app/main.py
```diff
@@ -1 +1,27 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> str | None:
+    if not command.startswith("mv "):
+        return "Wrong command"
+
+    _, source_file, new_file = command.split()
+    new_file = new_file.rstrip("/")
```
Dimosphen1, line 9: Do not use operation-system specific delimiters in the code, your solution should be platform-agnostic

app/main.py
```diff
@@ -1 +1,27 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> str | None:
+    if not command.startswith("mv "):
+        return "Wrong command"
+
+    _, source_file, new_file = command.split()
```
Dimosphen1, line 8: Make sure that command consists of exactly three elements before split or use
```suggestion
    _, source_file, new_file, *_ = command.split()
```

Dimosphen1 (changes requested):

app/main.py
liudmylaperchevska: it also works without 9th line

app/main.py
```diff
@@ -1 +1,26 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> str | None:
+    if not command.startswith("mv ") or len(command.split()) != 3:
+        return "Wrong command"
+
+    _, source_file, new_file, *_ = command.split()
```
pavlopro, line 8: Firstly, perform this operation and save the result into the variable 'result'. Then, you can check if the length of 'result' is 3 and if the first element is 'mv'. If so, unpack the result into variables.

pavlopro (changes requested):

PR #633, opened 2024-04-06
Solution

OttoLindholm:


app/main.py
```diff
@@ -1 +1,23 @@
...
+    command = command.split()
+
+    if len(command) != 3:
+        return
+
+    if command[0] != "mv":
+        return
+
+    if command[1] == command[2]:
+        return
```
Dimosphen1, line 14: Use unpacking to avoid using indexing and make code more descriptive

Dimosphen1, line 14: Join these two if conditions into one

app/main.py
```diff
@@ -1 +1,23 @@
...
+
+    second_file_dir = "/".join(command[2].split("/")[:-1])
```
Dimosphen1, line 16: Do not use any specific directory delimiters, your code should be platform-agnostic

Dimosphen1 (changes requested):

vasylhnatiuk (approved):

PR #634, opened 2024-04-07
Solution

SerhiiMar:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command, old_file_name, new_file_name = command.split()
```
Oleksl888, line 5: what if command is wrong and you will try to unpack less or more elements from a split?

app/main.py
```diff
@@ -1 +1,17 @@
...
+    if command != "mv":
+        return
+
+    if "/" in new_file_name:
```
Oleksl888, line 9: DIfferent OS use different slashes for separators. Use os.path.split to handle this

app/main.py
```diff
@@ -1 +1,17 @@
...
+        new_file_path = new_file_name.split("/")
+        new_file_path.pop(-1)
+        os.makedirs(os.path.join(*new_file_path), exist_ok=True)
+
+    with (open(old_file_name) as old, open(new_file_name, "w") as new):
```
Oleksl888, line 14: how do you handle errors when while working with file?

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+def move_file(command: str) -> None:
+    command = command.split()
+
+    if len(command) != 3:
+        raise ValueError(
+            f"unexpected number of values (expected 3, got {len(command)})"
+        )
+    if command[0] != "mv":
+        raise ValueError(f"{command[0]}: command not found")
```
Oleksl888, line 12: you can put it in one if statement and pput your logic under this if:
```
if len(command) == 3 and command[0] == "mv":
    rest of logic below

File: app/main.py [hunk 5]
```diff
@@ -1 +1,34 @@
...
+
+    old_file_name = command[1]
```
Oleksl888, line 14: if you pass previous check you can unpack command in one line

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+    if not os.path.exists(old_file_name):
+        raise FileNotFoundError(f"no such file '{old_file_name}'")
+
+    old_file_tail = os.path.split(old_file_name)[1]
+    new_file_tail = os.path.split(new_file_name)[1]
+    if not old_file_tail:
+        raise ValueError(f"'{old_file_name}' is not a file")
+    if not new_file_tail:
+        raise ValueError(f"'{new_file_name}' is not a file")
```
Oleksl888, line 25: this can be one if statement

app/main.py
```diff
@@ -1 +1,34 @@
...
+        raise ValueError(
+            f"unexpected number of values (expected 3, got {len(command)})"
+        )
+    if command[0] != "mv":
+        raise ValueError(f"{command[0]}: command not found")
+
+    old_file_name = command[1]
+    new_file_name = command[2]
+
+    if not os.path.exists(old_file_name):
```
Oleksl888, line 17: you can rather wrap your lines 31 and 32 in try except block which will cover more issues that can happen when working with file

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,31 @@
...
+        raise ValueError(f"'{command}' must contain two file names")
+
+    new_file_path = os.path.split(new_file_name)[0]
+    if new_file_path:
+        os.makedirs(new_file_path, exist_ok=True)
+
+    try:
+        with open(old_file_name) as old, open(new_file_name, "w") as new:
+            new.write(old.read())
+    except FileNotFoundError:
```
Oleksl888, line 24: OSError would be enough

app/main.py
```diff
@@ -1 +1,31 @@
...
+        raise ValueError(f"'{command}': invalid command input")
+
+    old_file_name, new_file_name = commands[1:]
+
+    old_file_tail = os.path.split(old_file_name)[1]
+    new_file_tail = os.path.split(new_file_name)[1]
+    if not old_file_tail or not new_file_tail:
+        raise ValueError(f"'{command}' must contain two file names")
+
+    new_file_path = os.path.split(new_file_name)[0]
```
Oleksl888, line 17: Check overall logic of your code - it fails when destination path is a directory:
`mv file.txt first_dir/second_dir/`

Oleksl888 (changes requested):

app/main.py
SerhiiMar: It handled on lines 12 to 15
Code raises `ValueError` if there is no filenames in commands but only path.

Oleksl888: > If name in destination path ends with / it must be considered as a directory. 

There shouldn't be a ValueError in this case. if you execute `mv file.txt first_dir/second_dir/` it assumes that you will move file.txt inside first_dir/second_dir/

Oleksl888 (changes requested):

Oleksl888 (approved):
Approved. 
You could simplify the logic by using `shutil.move` instead of `with open..` and `os.remove` to make your code cleaner + don't forget to add meaningful messages to all errors (line21)

PR #636, opened 2024-04-09
solution

CharmNiovi:


app/main.py
```diff
@@ -1 +1,31 @@
...
+
+    if extra_path:
+        raise ValueError(
+            "Expected 2 arguements"
+        )
+
+    if not os.path.exists(source_path):
+        raise FileExistsError(f"File {source_path} not exists")
+
+    if destination_dir and not os.path.exists(destination_dir):
```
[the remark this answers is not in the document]
CharmNiovi: > Also you anyway don't need to check `not os.path.exists(source_path)` twice. It will raise the error after the first check
If the file is not found, then directories will be created, and to prevent this, I check beforehand

app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+import shlex
+import os
+
+
+def move_file(command: str) -> None:
+    if command[:3] != "mv ":
```
Oleksl888, line 6: why not simplify the whole check process? 
You can split the file first (by the way why do you choose to use shlex.split() instead of built-in split()?)
Then you check if "mv" is there if len(command) is ok and put your main logic below

app/main.py
```diff
@@ -1 +1,31 @@
...
+            "Expected 2 arguements"
+        )
+
+    if not os.path.exists(source_path):
+        raise FileExistsError(f"File {source_path} not exists")
+
+    if destination_dir:
+        os.makedirs(destination_dir, exist_ok=True)
+
+    with (
```
Oleksl888, line 25: what if destination path will be passed like so `mv file.txt dir1/dir2/` ?

app/main.py
```diff
@@ -1 +1,31 @@
...
+        open(source_path, "r") as source_file,
+        open(destination_path, "w") as destination_file
```
Oleksl888, line 27: its good practice to wrap working with files with try-except to handle errors

Oleksl888 (changes requested):

app/main.py
CharmNiovi: @Oleksl888 
> why not simplify the whole check process? You can split the file first (by the way why do you choose to use shlex.split() instead of built-in split()?) Then you check if "mv" is there if len(command) is ok and put your main logic below

I chose shlex.split() because of the task of writing a similar command to the "mv" in bash, if there are spaces in the file, they can be escaped with \ or the full file name with spaces will be enclosed in quotes, for built-in split() you will need additional logic, and in shlex.split() it was already written

LiudmylaKulinchenko (approved):
Good job! 🌼

PR #639, opened 2024-04-10
solution

nrjss67:


app/main.py
```diff
@@ -1 +1,22 @@
...
+
+
+def move_file(command: str) -> None:
+    command_split = command.split()
+    if len(command_split) == 3 and command_split[0] == "mv":
+        comm, file_name, path = command_split
+
+    if "/" not in path:
+        os.rename(file_name, path)
+        return
```
vsmutok, line 11: it's better to raise some error in this case so that the user knows if something goes wrong

app/main.py
```diff
@@ -1 +1,22 @@
...
+
+    if path.endswith("/"):
+        os.makedirs(path, exist_ok=True)
+        os.replace(file_name, path)
+        return
```
vsmutok, line 16: I think this is redundant

app/main.py
```diff
@@ -1 +1,22 @@
...
+
+    if path.endswith(".txt"):
+        new_file_name = path.split("/")[-1]
+        path_without_fn = "/".join(path.split("/")[:-1])
+        os.makedirs(path_without_fn, exist_ok=True)
+        os.replace(file_name, f"{path_without_fn}/{new_file_name}")
```
vsmutok, line 22: why `.txt` ?

What if we need move other type of file? main.py for example

vsmutok (changes requested):

vsmutok (changes requested):
.

vasylhnatiuk (approved):

PR #640, opened 2024-04-10
Solution

sserkov84:


app/main.py
```diff
@@ -1 +1,20 @@
...
+    file1, file2 = c_args[1], c_args[2]
+    if "/" in c_args[2]:
+        splitter = []
+        path = c_args[2]
+        for num in range(len(path) - 1):
+            if path[num] == "/":
+                splitter.append(num)
+                print(path[:num])
+                if not os.path.isdir(path[:num]):
+                    os.mkdir(path[:num])
```
NazarSheremet, line 17: This part of the code can be simplified. Will be better to use the function `os.path.dirname()` to separate dir name from all path and `os.makedirs()` to create dir

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> bool:
+    c_args = command.split()
+    if command == "" or c_args[0] != "mv" or c_args[1] == c_args[2]:
+        return False
+    file1, file2 = c_args[1], c_args[2]
```
NazarSheremet, line 8: You can do this in `split()` function like 
`cmd, file1, file2 = command.split()`

NazarSheremet (changes requested):

YuriiSmolii (approved):
LGTM

PR #641, opened 2024-04-12
move_file_v1.0

YegorVolkov:
1) file moving aspect works
2) file renaming aspect works
3) proper cmd check implemented

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,20 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    cmd, file_1, file_2 = command.split()
+    if cmd != "mv":
+        print('only "mv" command is supported')
+        return None
+    if "/" not in file_2:
+        os.rename(file_1, file_2)
+    else:
+        dir_list = file_2.split("/")[0:-1]
+        os.mkdir(dir_list[0])
+        if len(dir_list) >= 2:
+            for i in range(len(dir_list[0:-1])):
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,20 @@
...
+        return None
+    if "/" not in file_2:
+        os.rename(file_1, file_2)
+    else:
+        dir_list = file_2.split("/")[0:-1]
+        os.mkdir(dir_list[0])
+        if len(dir_list) >= 2:
+            for i in range(len(dir_list[0:-1])):
+                os.mkdir(f'{"/".join(file_2.split("/")[0:i + 1])}'
+                         f"/{dir_list[i + 1]}")
```
NazarSheremet, line 17: This part of the code can be simplified. Will be better to use the function `os.path.dirname()` to separate dir name from all path and `os.makedirs()` to create all dir path

NazarSheremet (changes requested):

NazarSheremet:
> 4. I assume tht every task of Mate academy needs to update its requirements.txt
>    (flake8 is not up to date, which causes faulty-passing of the flake8 check on student's side)

I slightly misunderstood what you meant. All tasks are recommended to be performed within separate venvs, so the version and configuration of flake8 will be the same locally and on auto checks.

YegorVolkov:
> [NazarSheremet](/NazarSheremet)

I might have mistaken..

YuriiSmolii (approved):
Well done!

PR #647, opened 2024-04-16
Development

MikolaRozhansky:


app/main.py
```diff
@@ -1 +1,32 @@
...
+        cmd, source_name, destination_name = command.split()
+    except Exception:
+        print("Error in cmd!")
+    print(cmd, source_name, destination_name)
+
+    dir_str = ""
+
+    if cmd == "mv":
+        if "/" in destination_name:
+            list_dir = destination_name.split("/")
```
EdAlekseiev, line 16: Don't use `split` for working with directions, because each OS has personal separator. You should use `os.path.split`

app/main.py
```diff
@@ -1 +1,32 @@
...
+
+            for i in range(len(list_dir) - 1):
+                dir_str += list_dir[i] + "/"
+
+                try:
+                    print(dir_str)
+                    os.mkdir(dir_str)
+                except Exception:
+                    print("Err_dir")
```
EdAlekseiev, line 25: Use os.makedirs with exist_ok to avoid errors.
```suggestion
                os.makedirs(dir_str, exist_ok=True)
```

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,23 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    cmd, source_name, destination_name = "", "", ""
+    try:
+        cmd, source_name, destination_name = command.split()
+    except Exception:
+        print("Error in cmd!")
```
vsmutok, line 9: Also, use `print()` is not very appropriate here; for example, if something goes wrong, a message will be printed, and the program will continue its execution, which will cause an error in the subsequent stages. Therefore, if the command is incorrect, I advise to immediately terminate the function's action.

vsmutok (changes requested):

EdAlekseiev (approved):

PR #648, opened 2024-04-17
Solution

salaxambar:


app/main.py
```diff
@@ -1 +1,24 @@
...
+        if os.path.exists(first_file):
+            if "/" in second_file:
+                directories = second_file.split("/")
+                directories = directories[0:len(directories) - 1]
+                current = ""
+                for directory in directories:
+                    current += directory
+                    if not os.path.exists(current):
+                        os.mkdir(current)
+                    current += "/"
```
NazarSheremet, line 18: This part of the code can be simplified. Will be better to use the function os.path.dirname() to separate dir name from all path and os.makedirs() to create all dir path

app/main.py
```diff
@@ -1 +1,24 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    part = command.split()
+    first_file = part[1]
+    second_file = part[2]
```
NazarSheremet, line 7: You can simplify this code
```suggestion
    cmd, first_file, second_file  = command.split()
```

NazarSheremet (changes requested):

NazarSheremet (approved):
Good job!

PR #650, opened 2024-04-20
Develop

Nyxan:


tests/file.txt
```diff
@@ -0,0 +1,3 @@
+This is some
```
Oleksl888, line 1: file should not be in PR

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> bool:
+    if len(command.split()) != 3:
+        raise ValueError
+    cmd, source_file, new_file = command.split()
```
Oleksl888, line 7: do you check that command is actually 'mv'?

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> bool:
+    if len(command.split()) != 3:
+        raise ValueError
```
Oleksl888, line 6: what about error message for value error here?

app/main.py
```diff
@@ -1 +1,12 @@
...
+    cmd, source_file, new_file = command.split()
+    os.makedirs(os.path.dirname(os.path.abspath(new_file)),
```
Oleksl888, line 8: doesnt cover the case `mv file.txt first_dir/second_dir/`

app/main.py
```diff
@@ -1 +1,12 @@
...
+                exist_ok=True)
+    with open(source_file, "r") as file_in, open(new_file, "w") as file_out:
```
Oleksl888, line 10: this needs to be wrapped with try-except to catch any tracebacks related to file operations

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,26 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_split = command.split()
+    if len(command_split) != 3:
```
Oleksl888, line 6: You can simplify this like so:
```
if len(command_breakdown) == 3 and command_breakdown[0] == "mv":
  rest of logic here
OR
if len(command_breakdown) != 3 or command_breakdown[0] != "mv":
   raise ValueError..
rest of logic below

File: app/main.py [hunk 7]
```diff
@@ -1 +1,26 @@
...
+        raise ValueError(f"Invalid command: {command}, expected 3 arguments")
+    if command_split[0] != "mv":
+        raise ValueError("First argument must be 'mv'")
+    cmd, source_file, new_file = command_split
```
Oleksl888, line 10: since you don't use `cmd` you can mark it as `_`

app/main.py
```diff
@@ -1 +1,26 @@
...
+    if new_file.endswith("/"):
```
Oleksl888, line 11: this wouldnt work on all systems - use `os.path.split` instead

app/main.py
```diff
@@ -1 +1,26 @@
...
+        new_dir = new_file
+        new_file = os.path.join(new_dir, os.path.basename(source_file))
+    else:
+        new_dir = os.path.dirname(new_file)
+    os.makedirs(os.path.abspath(new_dir), exist_ok=True)
+    try:
+        with (open(source_file, "r") as file_in,
```
Oleksl888, line 18: there's too much going here and it's hard to read - try using shutil.move instead

app/main.py
```diff
@@ -1 +1,26 @@
...
+              open(os.path.join(
+                  new_dir,
+                  os.path.basename(new_file)
+              ), "w") as file_out):
+            file_out.write(file_in.read())
+        os.remove(source_file)
+    except OSError:
+        raise
```
Oleksl888, line 26: print some error message for user here

Oleksl888 (changes requested):

app/main.py
Nyxan: isn't it better to separate these errors

tests/file.txt
Oleksl888: Please remove. I insist

Oleksl888 (changes requested):
Very nice and clean solution. But please remove extra files from PR

Nyxan:
> Very nice and clean solution. But please remove extra files from PR

Sorry, I thought I deleted this file, now it's remove

Oleksl888 (approved):
Very Nice!

PR #653, opened 2024-04-30
Solution

rzaliznyi:


app/main.py
```diff
@@ -1 +1,23 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    parts = command.split()
+
+    source_file, destination_path = parts[1], parts[2]
```
Dimosphen1, line 7: Use unpacking and avoid indexing to retrieve values for better code style. You should check that command starts with `mv` as well

app/main.py
```diff
@@ -1 +1,23 @@
...
+
+    if os.path.exists(source_file):
+        if destination_path.endswith("/"):
```
Dimosphen1, line 10: Do not use any os-specific directory delimiters in the code, your code should be platform-agnostic

app/main.py
```diff
@@ -1 +1,23 @@
...
+            if not os.path.exists(destination_path):
+                os.makedirs(destination_path)
```
Dimosphen1, line 12: You could avoid using this if condition if you pass `exist_ok=True` flag to the `makedirs`

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,23 @@
...
+                                            os.path.basename(source_file))
+
+        else:
+            if os.path.isdir(destination_path):
+                destination_file = os.path.join(destination_path,
+                                                os.path.basename(source_file))
+            else:
+                destination_file = destination_path
+
+                os.replace(source_file, destination_file)
```
Dimosphen1, line 23: Command `move_file("mv file.txt first_dir/second_dir/third_dir/file2.txt")` does not work as expected in case directories are absent. Resolve this issue

app/main.py
```diff
@@ -1 +1,23 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    parts = command.split()
+
+    if parts[0] == "mv":
+        cmd, source_file, destination_path = parts
```
Dimosphen1, line 8: Avoid indexing, use unpacking and default return condition instead
```suggestion
    cmd, source_file, destination_path, *_ = command.split()

    if cmd != "mv":
        return
```

Dimosphen1 (changes requested):

EdAlekseiev (approved):

PR #658, opened 2024-05-01
Solution

Kate-Kotushenko:


app/main.py
```diff
@@ -1 +1,15 @@
...
+
+def move_file(command: str) -> None:
+    data = command.split()
+    if data[0] == "mv" and len(data) == 3:
+        dirs = data[2].split("/")
+        folder = dirs[0]
+        for i in dirs[1::]:
+            if not os.path.isdir(folder):
+                os.mkdir(folder)
+            folder += "/" + i
```
vsmutok, line 12: use `os.path.dirname` and `os.makedirs` instead of using a loop

vsmutok (changes requested):

README.md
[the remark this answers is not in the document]
Kate-Kotushenko: it's accidental

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    data = command.split()
+    if data[0] == "mv" and len(data) == 3 and data[1] != data[2]:
```
vsmutok, line 6:
```suggestion
    if len(data) == 3 and data[0] == "mv" and data[1] != data[2]:
```
first it is better to check that the length is three

app/main.py
```diff
@@ -1 +1,11 @@
...
+        if os.path.dirname(os.path.dirname(data[2])):
+            os.makedirs(os.path.dirname(data[2]))
```
vsmutok, line 8: Don't repeat yourself 
`os.path.dirname` - you do this three times

README.md
vsmutok, line 1: put this file back

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #666, opened 2024-05-14
MoveComplete

YakovK1808:


app/main.py
```diff
@@ -1 +1,24 @@
...
+        if os.path.exists(file_in):
+            if "/" in file_out:
+                catalogs = file_out.split("/")
+                catalogs = catalogs[0:len(catalogs) - 1]
+                current_catalog = ""
+                for catalog in catalogs:
+                    current_catalog += catalog
+                    if not os.path.exists(current_catalog):
+                        os.mkdir(current_catalog)
+                    current_catalog += "/"
```
NazarSheremet, line 18: Will be better to use the function os.path.dirname() to separate dir name from all path and os.makedirs() to create all dir path

NazarSheremet (changes requested):

YuriiSmolii (approved):
LGTM

PR #668, opened 2024-05-19
solution

haldaniko:


app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command = command.split()
+    cmd, old_loc, new_loc = command
```
vsmutok, line 6: Make sure you check that the `mv` input command is correct.
Also before unpacking, check that the length of `command ` is equal to thre

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #669, opened 2024-05-19
Hell yeah solution

RuslanLevchuk:


app/main.py
```diff
@@ -1 +1,20 @@
...
+        new_path = "/".join(new_path[:-1]) + "/"
+    if new_path:
+        if not os.path.exists(new_path):
+            os.makedirs(new_path)
+    with (
+        open(source_file_name, "r") as source_file,
+        open(new_path + new_file_name, "w") as new_file
+    ):
+        new_file.write(source_file.read())
+    os.remove(source_file_name)
```
vsmutok, line 20: 1. **Command validation**: Add validation to ensure the command line input is correct and follows the format `mv source destination`.

2. **Use os.path.join**: Use `os.path.join` to concatenate path parts to ensure cross-platform compatibility.

3. **Use os.makedirs**: The `os.makedirs` function can create all necessary directories, ignoring the error if the directory already exists when you pass the `exist_ok=True` parameter.

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> str | None:
+    command, source_file_name, new_path = command.split()
```
vasylhnatiuk, line 5: you should check if command has 3 arguments before it

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> str | None:
+    if len(command.split()) != 3:
+        return "Not enough arguments in command"
+    command, source_file_name, new_path = command.split()
```
EdAlekseiev, line 7: Don't use `command.split()` twice. You should create separate variable for storing the result of `command.split()` and use it where you need

app/main.py
```diff
@@ -1 +1,19 @@
...
+    if command != "mv":
+        return "Incorrect command"
+    new_file_name = new_path.split("/")[-1]
```
EdAlekseiev, line 10: Use os.path.split because each operating system has different directory separators.

EdAlekseiev (changes requested):

vasylhnatiuk (approved):

PR #670, opened 2024-05-20
Solution

taras-tkachuk:


app/main.py
```diff
@@ -1 +1,35 @@
...
+            os.mkdir(path)
+        path += "/"
+
+    if name:
+        # path = dir/new_name.txt
+        copy_file(old_file_name, path + name)
+    else:
+        # path = dir/
+        copy_file(old_file_name, path + old_file_name)
+    os.remove(old_file_name)
```
vsmutok, line 35: 1. **Command Validation**:
   - Validate the command to ensure it follows the expected format `mv source_file target_path`.

2. **Use `os.path.join`**:
   - Use `os.path.join` to concatenate path parts to ensure cross-platform compatibility.

3. **Simplify Directory Creation**:
   - Use `os.makedirs(path, exist_ok=True)` to create all necessary directories at once and avoid errors if directories already exist.

4. **Remove Unnecessary Checks**:
   - Simplify the check for whether the destination is a directory by directly using the path provided.

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+import os
+import shutil
+
+
+def move_file(command: str) -> None:
+    command_split = command.split(" ")
```
vasylhnatiuk, line 6:
```suggestion
    command_split = command.split()
```

vasylhnatiuk (changes requested):

PR #673, opened 2024-05-26
Solution

GGsosna:


app/main.py
```diff
@@ -1 +1,27 @@
-# write your code here
+import os
+import shutil
+
+from app.validator import validation_command
+
+
+def move_file(command: str) -> None:
+    import shlex
```
EdAlekseiev, line 8: Don't use `import` inside function. You should place all your imports on start of file

EdAlekseiev (changes requested):

GGsosna:
this code is better

app/main.py
```diff
@@ -1 +1,26 @@
...
+import os
+import shutil
+
+
+def move_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3 or parts[0] != "mv":
+        raise ValueError("Invalid command format. Use: mv source destination")
+
+    src = parts[1]
```
vasylhnatiuk, line 10:
```suggestion
    src, dest = parts[1:]
```

vasylhnatiuk (changes requested):

EdAlekseiev (approved):

PR #676, opened 2024-05-29
The solution of the task

Ja0ni:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command, file, direct = command.split(" ")
```
vsmutok, line 5:
```suggestion
    command, file, direct = command.split()
```

vsmutok, line 5: first check if the length of the split result is three, otherwise it will throw an error

app/main.py
```diff
@@ -1 +1,12 @@
...
+    if command == "mv":
+        new_direct = direct.split("/")
+        if len(new_direct) > 1:
+            os.makedirs("/".join(new_direct[:-1]), exist_ok=True)
```
vsmutok, line 9: use `os.path.dirname`  or `os.path.split`

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    action, file, direct = command.split()
```
vsmutok, line 5: first check if the length of the split result is three, otherwise it will throw an error

app/main.py
```diff
@@ -1 +1,13 @@
...
+    if action != "mv" or len(command.split()) != 3:
+        raise ValueError
```
vsmutok, line 7:
```suggestion
        raise ValueError(some message)
```

vsmutok (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #678, opened 2024-06-02
Solution

b4oody:


app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command, file_name_in, file_name_out, *_ = command.strip().split()
```
Dimosphen1, line 5: Before further processing, make sure that command is `mv` and `file_name_in` is not the same as `file_name_out`

Dimosphen1 (changes requested):

PR #682, opened 2024-06-04
solution

AntonKorinchuk:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    cmd, file_name, location = command.split()
+    dir_ = os.path.dirname(location)
+    if dir_:
+        os.makedirs(dir_, exist_ok=True)
```
DevSpaciX, line 8: you can try to use := operator here . 
It will be more cleaner imho , but this variant is good also

DevSpaciX (approved):

PR #683, opened 2024-06-05
'Solution'

AlexTarasov57:


app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_name, file, direct, *_ = command.split()
```
vasylhnatiuk, line 5: you should check if command has 3 arguments before it

app/main.py
```diff
@@ -1 +1,14 @@
...
+
+    if "/" in direct:
+        directory = direct.split("/")
```
vasylhnatiuk, line 8: it would be better to use `os.path.split`

vasylhnatiuk (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #685, opened 2024-06-10
created move file

Timur5050:


Timur5050:
I can explain why if in 18's line is so long:
- makedirs - if we create folders with a long path: “dir/dir1/dir2” - then the function will create folders in any case, even if there is already “dir” or even “dir1”
- and if we create 1 folder through makedirs “dir” and “dir” already exists - then there will be a flush

app/main.py
```diff
@@ -1 +1,24 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_name, file_to_move, new_file = command.split()
```
vsmutok, line 5: first you have to check if you can do the baking, i.e. if the length of `command.split()` is equal to three

app/main.py
```diff
@@ -1 +1,24 @@
...
+    if command_name == "mv":
+        whole_data = ""
+        with open(file_to_move) as file:
+            whole_data = file.read()
```
vsmutok, line 9: use only one `with`

app/main.py
```diff
@@ -1 +1,24 @@
...
+
+        os.remove(file_to_move)
+        path_to_new_file = new_file.split("/")
```
vsmutok, line 12: better to use `os.path.split`

app/main.py
```diff
@@ -1 +1,24 @@
...
+
+        combined_path = ""
+        for path in path_to_new_file[:-1]:
+            combined_path = os.path.join(combined_path, path)
```
vsmutok, line 16: better to use `os.path.split`

vsmutok (changes requested):

app/main.py
Timur5050: but I don't need to split my path here, I need to do the opposite, to join the final path

app/main.py
```diff
@@ -1 +1,22 @@
...
+
+            if (combined_path
+                    and len(path_to_new_file) > 1
+                    and not os.path.exists(path_to_new_file[-2])):
+                os.makedirs(combined_path)
+
+            with (open(file_to_move, "r") as file_in,
+                  open(new_file, "w") as file_out):
+                whole_data = file_in.read()
+                file_out.write(whole_data)
```
Dimosphen1, line 21: There is no need to create `whole_data` variable, pass `file_in.read()` directly to the `file_out.write`

app/main.py
```diff
@@ -1 +1,22 @@
...
+        if command_name == "mv":
+            path_to_new_file = os.path.split(new_file)
+            combined_path = ""
+            for path in path_to_new_file[:-1]:
+                combined_path = os.path.join(combined_path, path)
+
+            if (combined_path
+                    and len(path_to_new_file) > 1
+                    and not os.path.exists(path_to_new_file[-2])):
+                os.makedirs(combined_path)
```
Dimosphen1, line 16: Instead of performing extra checks, use `os.makedirs` with `exist_ok=True` flag

Dimosphen1 (changes requested):
Several minor changes were requested

Dimosphen1 (approved):
LGTM, well done!

PR #686, opened 2024-06-10
sol1

Unlie9:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    cmd, file_name, direction = command.split()
+    dir_ = direction
```
Oleg-s-s, line 6: Why you create additional var?

Oleg-s-s (changes requested):

file.txt
```diff
@@ -0,0 +1,3 @@
+This is some
+ content for
+ the file.
```
DevSpaciX, line 3: Do not push redundat files into Pull reqest : )

DevSpaciX (changes requested):

file.txt
Unlie9: XD

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    cmd, file_name, direction = command.split()
+    dir_ = os.path.dirname(direction)
+
+    if dir_:
+        os.makedirs(dir_, exist_ok=True)
```
DevSpaciX, line 9: You could use here operator warlus operator 

```
if dir_ := os.path.dirname(direction):
    os.makedirs(dir_, exist_ok=True)
```

That way you code will be a bit cleaner : )

DevSpaciX (changes requested):
[Read about warlus](https://realpython.com/python-walrus-operator/) operator that is really useful

DevSpaciX (approved):
Perfect !

PR #698, opened 2024-06-28
Solution

zave52:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    file1 = command.split(" ")[1]
+    file2_path = command.split(" ")[2]
```
NazarSheremet, line 6:
```suggestion
    commands_list, file_out, file_in = commands_str.split()
```

app/main.py
```diff
@@ -1 +1,16 @@
...
+
+    for i in range(file2_path.count("/")):
+        path = "/".join(file2_path.split("/")[:i + 1])
+        if os.path.exists(path):
+            continue
+        os.mkdir(path)
```
NazarSheremet, line 12: Will be better to use os module functions `os.path.split()` to separate dir name and  `os.makedirs()` to create all dir tree

NazarSheremet (changes requested):
Need to fix suggestions

YuriiSmolii (approved):
LGTM

PR #705, opened 2024-07-11
Solution

YerepSemyon:


app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command, file_out, file_in = command.split()
```
vasylhnatiuk, line 5: you should check if command has 3 arguments before this line

vasylhnatiuk (changes requested):

PR #709, opened 2024-07-20
Implement move_file function

anton-bilous:


app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+import os
+import shlex
+import shutil
```
MNDonut, line 3: let's don't use extra modules. the purpose of this task is to ensure that you know how to work with files without some unknown modules

MNDonut (changes requested):

app/main.py
```diff
@@ -1 +1,32 @@
...
+        command, source_path, destination_path = command.split()
+        if command != "mv":
+            raise ValueError("The command is not 'mv'")
+    except ValueError as e:
+        raise ValueError("The command is not valid") from e
+
+    if source_path == destination_path:
+        return
+
+    *source_path, source_file = source_path.split("/")
```
oliasymak24, line 15: Use os.path.split() instead

app/main.py
```diff
@@ -1 +1,32 @@
...
+def move_file(command: str) -> None:
+    try:
+        command, source_path, destination_path = command.split()
+        if command != "mv":
+            raise ValueError("The command is not 'mv'")
+    except ValueError as e:
+        raise ValueError("The command is not valid") from e
+
+    if source_path == destination_path:
+        return
```
oliasymak24, line 13: try not to use blank return here, make a backward condition

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #710, opened 2024-07-21
Implemented main

OleksiukStepan:


app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    parsed_command = command.split()
```
Oleksl888, line 5: you should check if the command is valid - check that it has 3 elements and contains `mv`

app/main.py
```diff
@@ -1 +1,19 @@
...
+    path = parsed_command[-1].split("/")
```
Oleksl888, line 6: 1. instead of using indexes, better use unpacking.
2. `os.path.split()` will be better here since it will correctly split on all systems

app/main.py
```diff
@@ -1 +1,19 @@
...
+    directory = ""
+
+    for i in range(len(path) - 1):
+        directory += path[i]
+        if not os.path.exists(directory):
+            os.mkdir(directory)
+        directory += "/"
```
Oleksl888, line 13: use `os.path.join`  to build correct path for a file

app/main.py
```diff
@@ -1 +1,19 @@
...
+
+def move_file(command: str) -> None:
+    parsed_command = command.split()
+    path = parsed_command[-1].split("/")
+    directory = ""
+
+    for i in range(len(path) - 1):
+        directory += path[i]
+        if not os.path.exists(directory):
+            os.mkdir(directory)
```
Oleksl888, line 12: you can use `mkdirs` to create directories inside directories

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    parsed_command = command.split()
+    path = parsed_command[-1].split("/")
+    directory = ""
+
+    for i in range(len(path) - 1):
```
Oleksl888, line 9: Your code does not handle case when destination path is directory: `mv file.txt dir1/dir2/`

app/main.py
```diff
@@ -1 +1,19 @@
...
+        directory += path[i]
+        if not os.path.exists(directory):
+            os.mkdir(directory)
+        directory += "/"
+    directory += path[-1]
+
+    with open(parsed_command[1], "r") as f:
```
Oleksl888, line 16: `f` and `c` are not very obvious names for file handles. also you can use tools from `shutil` lib or `pathlib` to move files

Oleksl888, line 16: you should also add error handling here - there's things that can go wrong when working with files: there can be mistake in name, file can be opened by other user etc

Oleksl888 (changes requested):

app/main.py
OleksiukStepan: I am not using os.path.join in next commit, because i already have got correct directory format with os.path.split(), but i understand your hint. Thank you.

app/main.py
```diff
@@ -1 +1,21 @@
...
+
+def move_file(command: str) -> None:
+    try:
+        assert len(command.split()) == 3, "Command should have 3 arguments"
+        assert command.split()[0] == "mv", "First command should be 'mv'"
+    except AssertionError as e:
+        print(e)
+        return
+
+    order, old_file, new_location = command.split()
```
oliasymak24, line 13: Better to put in to the begging of the function and use in this assert `assert command.split()[0] == "mv"`

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #711, opened 2024-07-22
Implemented solution

dimak20:


app/main.py
```diff
@@ -1 +1,25 @@
...
+    if file == new_file:
+        return
+    if "/" not in new_file:
+        try:
+            shutil.move(file, new_file)
+        except FileNotFoundError:
+            print(f"{file} does not exist")
+        finally:
+            return
+    end_path = new_file.split("/")
```
oliasymak24, line 16: Use os.path.split() instead

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1 +1,25 @@
-# write your code here
+import os
+import shutil
+
+
+def move_file(command: str) -> None:
+    _, file, new_file = command.split()
+    if file == new_file:
+        return
```
oliasymak24, line 8: try not to use blank return here, make a backward condition

app/main.py
```diff
@@ -1 +1,25 @@
...
+            shutil.move(file, new_file)
+        except FileNotFoundError:
+            print(f"{file} does not exist")
+        finally:
+            return
+    end_path = os.path.split(new_file)
+    directory = end_path[:-1]
+    file_name = end_path[-1]
+    path_for_directory = os.path.join(*directory)
+    path_for_file = os.path.join(*directory, file_name)
```
oliasymak24, line 20: What is the differene between new_file and path_for_file?

oliasymak24 (changes requested):

app/main.py
dimak20: I tried considering the case when I recieve path with "/" but need to move/create file in another OS where other type of paths are used

MNDonut (approved):

PR #715, opened 2024-07-26
Solution

DSahalatyi:


app/main.py
```diff
@@ -1 +1,28 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command, file_name, move_to = command.split()
+    directories = []
+
+    if "/" in move_to:
+        move_to = move_to.split("/")
```
oliasymak24, line 9: Is ur code cross-platform ?) For example, will it work on win with \ ? Use os.path.split() instead

app/main.py
```diff
@@ -1 +1,28 @@
...
+        directories = move_to[:-1]
+        move_to = move_to[-1]
+
+    path = ""
+    for item in directories:
+        path += item
+        if not os.path.exists(path):
+            os.mkdir(path)
```
oliasymak24, line 17: You can use makedirs method instead

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #720, opened 2024-08-03
implement move_file

Morphin20th:


app/main.py
```diff
@@ -1 +1,17 @@
...
+
+
+def move_file(command: str) -> None:
+    split_command = command.split()
+
+    if split_command[0] != "mv":
+        return
+
+    source = split_command[1]
+    destination = split_command[2]
```
vsmutok, line 11: first check if the command length is three

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    split_command = command.split()
+
+    if split_command[0] != "mv":
+        return
```
vsmutok, line 8: it is better to print some informative message before that, so that the user understands what went wrong

vsmutok (changes requested):

PR #721, opened 2024-08-04
Solution

bknyrik:


app/main.py
```diff
@@ -1 +1,28 @@
...
+            os.makedirs(destination_dir_path)
+
+        with (
+            open(source_path) as source_file,
+            open(destination_path, "w") as destination_file
+        ):
+            for line in source_file:
+                destination_file.write(line)
+
+        os.remove(source_path)
```
vsmutok, line 28: 1. `command.split()` - do it only once
2. `endswith(".txt")` why do you need this check? this code should work with files of different extensions
3. try not to use a loop in this task

EdAlekseiev (approved):

PR #724, opened 2024-08-06
Solution Politov

Alexandr-Politov:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
```
Oleksl888, line 4: Nicely done :)
Can you do this exercise not using os.renames?
In this task we expect:
1. Check if command is correct (by the way not done in your implementation)
2. Check if destination is a file or dir and build correct path using methods `os.path.split, os.path.join, os.makedirs.`
3. Move file using `with open` and then removing the source file.
4. Your script must give the same result when commands are like so: `mv file.txt first_dir/second_dir/file.txt` and `mv file.txt first_dir/second_dir/`

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #726, opened 2024-08-10
Solution

Bogdan-Rylov:


app/main.py
```diff
@@ -1 +1,33 @@
...
+
+def move_file(command: str) -> None:
+    try:
+        command, source_path, destination_path = command.split()
+    except ValueError:
+        print("Wrong command format")
+    else:
+        if command == "mv" and source_path != destination_path:
+            if not os.path.isfile(source_path):
+                raise ValueError(f"File '{source_path}' doesn't exist")
```
Oleksl888, line 13: You can simplify your code but not checking source path here but instead wrap the block of code where you read/write the file in try except - this will allow to process all file-related error including wrong path

app/main.py
```diff
@@ -1 +1,33 @@
...
+            source_path = os.path.join(*re.split(r"[\\/]", source_path))
```
Oleksl888, line 14: os.path.split() is prefereable method to correctly split file path on all systems

app/main.py
```diff
@@ -1 +1,33 @@
...
+            destination_path = os.path.join(
+                *re.split(r"[\\/]", destination_path)
+            )
+            if (
+                    source_path.endswith(os.sep)
+                    or destination_path.endswith(os.sep)
```
Oleksl888, line 20: this block of code is unnecessary if you use try-except when you work with file.
also  `move_file("mv file1.txt first_dir/second_dir/third_dir/")`  is a valid command. Your program should move file when destination is a directory as well

app/main.py
```diff
@@ -1 +1,33 @@
...
+            ):
+                return
+
+            if os.sep in destination_path:
```
Oleksl888, line 24: using `os.path.isdir(destination_path)` would be a preferable way to check if path is a dir or not

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #731, opened 2024-08-16
Solution

juliazelena:


LLkaia (approved):
Gj!
But in general, try to write simpler code. At least, instead of using `if` nested in `else`, you can use `elif`.

PR #732, opened 2024-08-17
My solution

lavryniuk-dev:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+from os import path, makedirs
+from shutil import move
+
+
+def move_file(command: str) -> None:
+    _, source, destination = command.split()
+
+    if (not (command.startswith("mv "))
```
LiudmylaKulinchenko, line 8: Why don't you get this in the line 6, where you split str if you want to check it here?

LiudmylaKulinchenko (approved):
Looks good 🙂

app/main.py
lavryniuk-dev: I thought that it would work faster

PR #745, opened 2024-08-30
Solution

Alex414414:


app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    file_names = command.split()[1:]
```
vsmutok, line 5: you should check if command has 3 arguments before it

app/main.py
```diff
@@ -1 +1,21 @@
...
+        second_file_path.pop()
+        path_to_second = "/".join(second_file_path)
+        if not os.path.exists(path_to_second):
+            os.makedirs(path_to_second, exist_ok=True)
+        with (
+            open(file_names[1], "w") as file_2,
+            open(file_names[0], "r") as file_1
+        ):
+            file_2.write(file_1.read())
+        os.remove(file_names[0])
```
vsmutok, line 21: Optimize path checks:
Use the os.path or pathlib module for working with paths.

vsmutok (changes requested):

PR #746, opened 2024-09-03
Solution, tests passed, flake8 - ok

samuel-edmund-morgan:


app/main.py
```diff
@@ -1 +1,44 @@
...
+        raise InvalidPathError("Invalid path for moving file")
+    elif not args_list[1].split("/")[-1]:
+        raise InvalidFileError("Invalid file name")
+    else:
+        dirs_list.extend(args_list[2].split("/"))
+        if dirs_list[-1]:
+            new_file_name = dirs_list[-1]
+            del dirs_list[-1]
+        else:
+            new_file_name = args_list[1].split("/")[-1]
```
NazarSheremet, line 27: Simplify your solution using `os.path.dirname() ` function. It returns dir name from the path

NazarSheremet (changes requested):

NazarSheremet (approved):

PR #747, opened 2024-09-03
Solution

razumovskayada:


app/main.py
```diff
@@ -1 +1,29 @@
...
+                if not os.path.exists(make_direction):
+                    os.mkdir(make_direction)
+        path = os.path.join(make_direction, path_directions[-1])
+        with (open(command_list[1], "r") as file_in,
+                open(path, "w") as file_out):
+            for line in file_in:
+                file_out.write(line)
+        os.remove(command_list[1])
+    elif command_list[0] == "mv" and len(command_list) == 3:
+        with (open(command_list[1], "r") as file_in,
```
vasylhnatiuk, line 25: DRY, avoid duplicating

app/main.py
```diff
@@ -1 +1,29 @@
...
+            and len(command_list) == 3
+            and "/" in command_list[2]):
+        path_directions = command_list[2].split("/")
+        make_direction = path_directions[0]
+        if not os.path.exists(make_direction):
+            os.mkdir(make_direction)
+        if len(path_directions) > 2:
+            for i in range(1, len(path_directions) - 1):
+                make_direction += "/" + path_directions[i]
+                if not os.path.exists(make_direction):
```
vasylhnatiuk, line 16: DRY, avoid duplicating

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):

PR #748, opened 2024-09-04
solution

uzlss:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+
+    _, file_name, copy_path = command.split()
```
vasylhnatiuk, line 6: you have to check if command has 3 arguments before it

vasylhnatiuk (changes requested):

PR #749, opened 2024-09-06
Solution

Deadmarvald:


app/main.py
```diff
@@ -1 +1,27 @@
...
+import os
+
+
+def move_file(command: str) -> None:
+    command_list = command.split(" ")
+    if command_list[0] != "mv" or len(command_list) < 2:
+        return
+
+    source_file_name = command_list[1]
+    list_of_dirs = command_list[2].split("/")
```
NazarSheremet, line 10: You can assign and separate dir names this way 
```suggestion
    cmd, source_file_name, destination_path =  command_list
    destination_dir = os.path.dirname(destination_path)
```

NazarSheremet (changes requested):

app/main.py
```diff
@@ -1 +1,21 @@
...
+import os
+
+
+def move_file(command: str) -> None:
+    command_list = command.split(" ")
+    cmd, source_file_name, destination_path = command_list
+    destination_dir = os.path.dirname(destination_path)
+
+    if cmd != "mv" or len(command_list) < 2:
+        return
```
NazarSheremet, line 10: If the command contain more than 3, it's incorrect too 
```suggestion
    if cmd != "mv" or len(command_list) == 3:
        return
```

NazarSheremet (changes requested):

NazarSheremet (approved):
Good job!

PR #753, opened 2024-09-12
Solution

CheshireKate:


app/main.py
```diff
@@ -1 +1,18 @@
...
+        print("The format is incorrect")
+        return
+    source, destination = split_command[1], split_command[2]
+    parts = destination.split("/")
+    file_name = parts.pop()
+    current_path = ""
+    for part in parts:
+        current_path = os.path.join(current_path, part)
+        if not os.path.exists(current_path):
+            os.mkdir(current_path)
```
Dimosphen1, line 16: Use `exist_ok=True` flag here to avoid this if condition

app/main.py
```diff
@@ -1 +1,18 @@
...
+import os
+
+
+def move_file(command: str) -> None:
+    split_command = command.split()
+    if len(split_command) != 3 or split_command[0] != "mv":
+        print("The format is incorrect")
+        return
+    source, destination = split_command[1], split_command[2]
+    parts = destination.split("/")
```
Dimosphen1, line 10: Use the built-in methods to extract required values from the path. You code should be platform-agnostic and it should work with different directories delimiters without hardcoding them

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    split_command = command.split()
+    if len(split_command) != 3 or split_command[0] != "mv":
+        print("The format is incorrect")
+        return
+    source, destination = split_command[1], split_command[2]
```
Dimosphen1, line 9: Use unpacking here

Dimosphen1 (changes requested):

PR #756, opened 2024-09-18
solution for move file

dissom:


app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    mv, source_file, result = command.split()
```
vasylhnatiuk, line 5: you should check if command has 3 arguments before this line

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #769, opened 2024-10-04
Created universal move_file func

Mxbely:


app/main.py
```diff
@@ -1 +1,32 @@
...
+import os
+
+
+def move_file(command: str) -> None:
+    command = command.split()
+    route = command[2].split("/")
+    if command[0] != "mv":
+        return
+    if len(route) > 1:
+        new_file_name = route[-1]
```
Atikiho, line 10: You don't need to declare new_file_name here

app/main.py
```diff
@@ -1 +1,32 @@
...
+            os.mkdir(directory)
+        os.chdir(directory)
+    for _ in range(len(route)):
+        os.chdir("..")
+    os.rename(command[1], command[2])
+
+
+if __name__ == "__main__":
+    path = "mv file.txt file2.txt"
+    move_file(path)
```
Atikiho, line 32: You should remove your checks before pushing solution

Atikiho (changes requested):

PR #772, opened 2024-10-07
Solution

KostiantinPastukh:


app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_name, file_name, file_path = command.split()
```
Dimosphen1, line 5: Please, make sure that unpacking would not fail in case `split` returns more than three arguments
```suggestion
    command_name, file_name, file_path, *_ = command.split()
```

app/main.py
```diff
@@ -1 +1,14 @@
...
+
+    if command_name != "mv" or file_path[-1] == "/":
```
Dimosphen1, line 7: Do not use directory-specific delimiters in the code, your code should be platform-agnostic

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_name, file_name, file_path, *_ = command.split()
```
Dimosphen1, line 5: The `file_path` argument name is incorrect, you could use `file_name_out` or similar name to indicate that the variable's content is not only a path, but a filename as well

app/main.py
```diff
@@ -1 +1,14 @@
...
+
+    if command_name != "mv":
```
Dimosphen1, line 7: Please, checks that file names are not the same as well

Dimosphen1 (changes requested):
Several minor changes were requested

PR #813, opened 2024-11-11
Solution

88victory88:


Alkaponees (changes requested):
Recommend add checking on length of command and if command start with mv

app/main.py
```diff
@@ -1 +1,27 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_list = command.split(" ")
+    first_command = command_list[0]
+    if first_command == "mv" and len(command) >= 6:
```
NazarSheremet, line 7: This check `len(command) >= 6` looks useless, need to check if command contains 3 parts

NazarSheremet (changes requested):

app/main.py
```diff
@@ -1 +1,27 @@
...
+import os
+
+
+def move_file(command: str) -> None:
+    command_list = command.split(" ")
+    first_command = command_list[0]
+    if first_command == "mv" and len(command_list) == 3:
+
+        file_old = command_list[1]
+        destination = command_list[2]
```
YuriiSmolii, line 10:
```suggestion
        file_old, destination = command_list[1], command_list[2]
```

YuriiSmolii (approved):
LGTM but could be simplified and refactored.

PR #838, opened 2024-12-14
Created context manager

shafran-andriy:


app/main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+import os
+
+from contextlib import contextmanager
+
+
+@contextmanager
+def move_file(command: str) -> None:
+    command_list = command.split()
+    if command_list[0] == "mv" and len(command_list) == 3:
```
pavlopro, line 9: first of all check that length is 3

pavlopro (changes requested):

app/main.py
shafran-andriy: yes it is included already, please recheck again

shafran-andriy: len(command_list) == 3

pavlopro: It should be on the first place 
`if len(command_list) == 3 and...:`

Imagine if command_list is an empty list. In the case where `command_list[0] == "mv"` is on the first place, you will encounter an error.

pavlopro (changes requested):

pavlopro (approved):

PR #841, opened 2024-12-15
Solution

IvankaKuzin:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command_list = command.split(" ")
+    index_new_directories = command_list[2].rfind("/")
```
4ikkamony, line 6: Got a problem here - you're getting last index, where "/" is found. But when there's no "/" it returns -1.
For example, 'mv file.txt file2.txt', there's no /, so index is -1

app/main.py
```diff
@@ -1 +1,18 @@
...
+    new_directories = command_list[2][:index_new_directories]
```
4ikkamony, line 7: Then, when you use this index in the slice, the last symbol gets cut off, 
with command 'mv file.txt file2.txt',
new_directories == a2.tx.

app/main.py
```diff
@@ -1 +1,18 @@
...
+    if command_list[0] == "mv":
+        os.makedirs(new_directories, exist_ok=True)
```
4ikkamony, line 9: thus, a directory "file2.tx" is created here. You may have noticed it is left behind after running pytest

4ikkamony (changes requested):
Nice work, you just need to check if the command you receive is valid and make sure the path is formatted properly, before creating directories

4ikkamony (approved):
Nice

PR #848, opened 2024-12-19
Solution

taraskhom:


app/main.py
```diff
@@ -1 +1,25 @@
...
+
+def move_file(command: str) -> None:
+
+    parts = command.split()
+    if len(parts) != 3 or parts[0] != "mv":
+        raise ValueError
+
+    source_file = parts[1]
+    destination = parts[2]
+
```
L1nk3rrr, line 13: when you split, already use an unpacking
`var1, var2, var3 = string.split()`

app/main.py
```diff
@@ -1 +1,25 @@
...
+    if destination.endswith("/"):
+        os.makedirs(destination, exist_ok=True)
+        destination = os.path.join(destination, os.path.basename(source_file))
+    else:
+        dest_dir = os.path.dirname(destination)
+        if dest_dir:
+            os.makedirs(dest_dir, exist_ok=True)
+
+    shutil.move(source_file, destination)
```
L1nk3rrr, line 25: can you do this action without shutil module. Try to do it with some basics, open files and os to remove file if needed

L1nk3rrr (changes requested):
Great job! Some points to do you need for finish this work

L1nk3rrr (approved):
Well done!

PR #869, opened 2025-01-06
Solution

whatmode:


whatmode (commented):
AI mentor didn't understand that this logic is correct according to option type

PR #888, opened 2025-02-03
Complete function move file.

kram3ko:


app/main.py
```diff
@@ -1 +1,32 @@
...
+        return
+
+    cmd, src, dest = data
+    src_file = Path(src)
+    dest_file = Path(dest)
+    dest_file.parent.mkdir(parents=True, exist_ok=True)
+
+    try:
+        if cmd != "mv":
+            raise ValueError(f"command: -->> {cmd} <<-- Allowed only: mv")
```
yarzhuravlov, line 17: Good job!

But a small question. Why cmd validation is performed after the destination folder is created?

kram3ko: Cause like, if could me more comands ;))  there like prepeared code ;) instead of data[0] != "mv

yarzhuravlov: But if the command is incorrect, there is no need to create folders because we don't know what exactly the user wants to do. Why not validate cmd before creating folders for destination?

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+from pathlib import Path
+
+
+def move_file(command_file1_file2: str) -> None:
+    data = command_file1_file2.split()
+    if len(data) != 3 and data[0] != "cmd":
```
Baranotik15, line 6: I wonder if it would be correct to use qr to check both conditions
if len(data) != 3 or data[0] != "cmd":
now the conditions are not met if 1 of the errors is present and the 2nd one is not

Baranotik15 (commented):
Good job!

Baranotik15 (changes requested):
correct your mistake and good job

Baranotik15 (approved):

PR #890, opened 2025-02-04
solution

Bleit1:


tests/test_main.py
```diff
@@ -1,8 +1,6 @@
 import os
 import shutil
-
 import pytest
-
```
L1nk3rrr, line 5: why did you remove empty lines here?

app/main.py
```diff
@@ -1 +1,53 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> [None, str]:
+    command_split = command.split(" ")
+    action = command_split[0]
+    source_file_option = command_split[1]
+    target_file_option = command_split[2]
```
L1nk3rrr, line 8: you can do it in one liner

app/main.py
```diff
@@ -1 +1,53 @@
...
+    target_type = None
+    target_folders_path = None
+
+    if action != "mv":
+        print(f"Wrong action is used{action}. mv action is expected")
```
L1nk3rrr, line 13: so if action not mv you will continue your code? Maybe we need to interrupt it?

app/main.py
```diff
@@ -1 +1,53 @@
...
+
+    if "/" in target_file_option:
+        target_folders_path = target_file_option.rsplit("/", 1)[0]
+        target_type = "folder"
+
+    if "/" not in target_file_option:
+        target_folders_path = target_file_option
+        target_type = "file"
```
L1nk3rrr, line 21: you can use 
`os.path.isdir(path)` or `os.path.isfile(path)` to know what it is

L1nk3rrr (changes requested):
Minor points for improvement, also don't forget about empty lines at end of the file and check git actions, if it passed all tests and flake, before send to review

PR #908, opened 2025-02-20
Solution

Nikita-pyth:


app/main.py
oliasymak24, line 1: Try to split your function into more functions

app/main.py
```diff
@@ -1 +1,25 @@
...
+import os
+
+
+def move_file(command: str) -> None:
+    command, from_f, to_f = command.split(" ")
+    directories = to_f.split("/")
+
+    if to_f[-1] == "/":
+        directories = directories[:-1]
+        file_name_moved = from_f.split("/")[-1]
```
oliasymak24, line 10: Is ur code cross-platform ?) For example, will it work on win with \ ? Use os.path.split() instead

app/main.py
```diff
@@ -1 +1,25 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    command, from_f, to_f = command.split(" ")
```
oliasymak24, line 5: no need to use space inside split method, it uses space by default

app/main.py
```diff
@@ -1 +1,25 @@
...
+        directories = directories[:-1]
+        file_name_moved = from_f.split("/")[-1]
+    else:
+        directories, file_name_moved = directories[:-1], directories[-1]
+
+    current_dir = ""
+    for directory in directories:
+        current_dir = os.path.join(current_dir, directory)
+        if not os.path.exists(current_dir):
+            os.mkdir(current_dir)
```
oliasymak24, line 18: u can use makedirs argument "exist_ok" here

app/main.py
```diff
@@ -1 +1,25 @@
...
+
+    new_file_path = os.path.join(current_dir, file_name_moved)
+    with open(from_f, "r") as src_file:
+        with open(new_file_path, "w") as dst_file:
```
oliasymak24, line 22: No need to use “with” twice here, u can make this using only once

app/main.py
oliasymak24, line 18: Try to use makedirs method

oliasymak24 (changes requested):

app/main.py
Nikita-pyth: The length of updated code is way shorter than it was before. Is there still a need of pliting it into smaller functions? Thx

app/main.py
```diff
@@ -1 +1,20 @@
...
+
+def move_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3:
+        raise ValueError("Command must be in the format: 'move from_f to_f'")
+
+    _, from_f, to_f = parts
+    current_dir = os.path.dirname(to_f)
+
+    if current_dir and not os.path.exists(current_dir):
```
oliasymak24, line 12: If you add `"exist_ok"` then you don't need this `if`.

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #949, opened 2025-03-31
Solution

olek007:


olek007:
@mateacademy-ai-mentor 
The code does check if the input string contains exactly three elements separated by spaces. If not it will raise error which will be handle in except block.
Changed to using os.path.split from .split

PR #1018, opened 2025-07-03
Implement file move functionality with directory creation

viannik:


DanilHamster (approved):

viannik:
@mateacademy-ai-mentor `if dest_dir: is already checking if the path exists, it will not perform folder creation then. And if you pass “.” to os.makedirs, the error will not occur. In Python, os.makedirs(“.”, exist_ok=True) simply checks the existence of the current directory and does not create anything, since it already exists.

PR #1020, opened 2025-07-07
solution for py-move-file

Hetmarum:


viannik (approved):
The current code does not check if new_path ends with a slash / and does not handle this case.
The code does not use os.path.join when constructing the destination path if the destination ends with a slash.

PR #1220, opened 2026-02-11
Create move_file func

danylobohomazov:


app/main.py
```diff
@@ -1 +1,25 @@
...
+    for index in range(len(dirs)):
+        try:
+            os.mkdir("/".join(dirs[:index + 1]))
+        except Exception:
+            continue
+    try:
+        with (open(file_name1, "r") as source_file,
+              open(file_name2, "w") as dest_file):
+            for line in source_file:
+                dest_file.write(line)
```
vsmutok, line 22: Try not to use a loop here.

app/main.py
```diff
@@ -1 +1,25 @@
...
+    if len(parts) != 3:
+        return
+    command_str, file_name1, file_name2 = parts
+    if file_name1 == file_name2 or command_str != "mv":
+        return
+    dirs = file_name2.split("/")
+    dirs.pop(-1)
+    for index in range(len(dirs)):
+        try:
+            os.mkdir("/".join(dirs[:index + 1]))
```
vsmutok, line 15: It’s better to use `makedirs` in this case.

app/main.py
```diff
@@ -1 +1,25 @@
...
+    parts = command.split()
+    if len(parts) != 3:
+        return
+    command_str, file_name1, file_name2 = parts
+    if file_name1 == file_name2 or command_str != "mv":
+        return
+    dirs = file_name2.split("/")
+    dirs.pop(-1)
+    for index in range(len(dirs)):
+        try:
```
vsmutok, line 14: use `os.path.join(...) `

vsmutok (changes requested):

PR #1250, opened 2026-03-17
Solution

namaskar108:


app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+import os
+
+
+def move_file(command: str) -> None:
+    files = command.split()
+    if len(files) != 3:
+        return
+
+    mv, file_1, file_2 = files
```
viktoria-rybenchuk, line 9: write descrptive variable names, like source_file and destination

app/main.py
```diff
@@ -1 +1,22 @@
...
+
+    if mv != "mv":
+        return
+
```
viktoria-rybenchuk, line 13: `file_name, directory = os.path.split(destination_path)` - it will be good to divide it like this, and then check if directory and then  use makedirs

viktoria-rybenchuk (changes requested):

PR #1265, opened 2026-03-31
Implemented solution for py-move-file

Neyrony:


Neyrony:
I can't use os.path.join() because of compatibility. I work on Windows and test was written for Linux so this comand make my code for Windows and my program cann't pass the test. The other thing i've fixed

PR #1267, opened 2026-04-01
solved

pawunder1992:


app/main.py
```diff
@@ -1 +1,18 @@
...
+import os
+
+
+def move_file(command: str) -> None:
+    curr_file = command.split()
+    if len(curr_file) != 3 and curr_file[0] != "mv":
+        return
+    old_file, new_file = curr_file[1], curr_file[2]
+    if new_file.find("/") != -1:
+        path = os.path.join(*new_file[:new_file.rfind("/")].split("/"))
```
AnyoneClown: Try to extract folder name via `os.path.dirname()`, it's gonna be a lot easier and more safe

app/main.py
```diff
@@ -1 +1,18 @@
...
+        try:
+            os.makedirs(path)
+        except FileExistsError:
+            with open(old_file) as old, open(new_file, "w") as new:
+                new.write(old.read())
```
AnyoneClown: Avoid try/except logic here.

os.makedirs(path, exist_ok=True) -> pass `exist_ok=True` argument and refactor your code

AnyoneClown (changes requested):

app/main.py
pawunder1992: ok, i make it because in mate academy tests i saw  exist_ok  must be False, my mistake

app/main.py
pawunder1992: now i know it, but in that time dont know what the method dirname and create it myself))

VadimPusiak788 (approved):

PR #1275, opened 2026-04-06
'Solution'

artradkevych:


app/main.py
```diff
@@ -1 +1,16 @@
...
+        return
+    operator, source, destination = command_list
+    if operator != "mv":
+        return
+    if destination.endswith("/"):
+        destination = os.path.join(destination, os.path.basename(source))
+    dir_path = os.path.dirname(destination)
+    if dir_path:
+        os.makedirs(dir_path, exist_ok=True)
+    os.rename(source, destination)
```
petrykivd, line 16: os.rename() will raise FileExistsError on Windows if destination already exists. Prefer os.replace() — it's atomic and cross-platform safe 😉

petrykivd (approved):
Good job! Check the comment and keep it in mind for the future 😊

PR #1281, opened 2026-04-11
Implemented move file function

PavelPereverzev1:


alisahoys (approved):
I think using binary mode is unnecessary

PR #1286, opened 2026-04-14
Sulution

Dmytro234:


Losiev (changes requested):
Check AI suggestions. There are few task items ignored

PR #1365, opened 2026-07-08
Solution

Teashrock:


Teashrock:
@mateacademy-ai-mentor 
I've fixed everything. Also, I used `os.makedirs` instead of multiple `os.mkdir`s because the former is efficient, and the latter is redundant code.

PR #1372, opened 2026-07-12
feat: implemented move_file() func

neeeksa:


app/main.py
```diff
@@ -1 +1,28 @@
...
+
+    try:
+        directory = os.path.dirname(new_file)
+
+        if directory:
+            os.makedirs(directory, exist_ok=True)
+
+        with open(origin_file_name, "r") as origin, open(new_file, "w") as new:
+            for line in origin:
+                new.write(line)
```
vsmutok, line 24:
```suggestion
        new_file.write(origin_file_name.read())
```
no need ti use loop here

vsmutok (changes requested):

